PR #60, opened 2023-03-21
Add simulation object

### LINKED ISSUE #15: Implement abstraction for TraCI Objects
## User stories
As a developer I want an abstraction over TraCI objects in order to have a more pythonic API.

## Dependencies
Depends on: 
 - #5 

## (Wiki) References
- <https://www.codementor.io/blog/pythonic-code-6yxqdoktzt>

## Acceptance Criteria
- [x] An object knows its own underlying TraCI id(s)
- [x] Common TraCI methods are implemented here
- [x] Decide who is responsible for updating the `SimulationObject`s

SaturnHafen:
Fixes #15 

## PR checklist

- [ ] ~~Breaking changes anounced~~


.github/workflows/python-app.yml
```diff
@@ -48,6 +48,10 @@ jobs:
       uses: actions/setup-python@v3
       with:
         python-version: "3.10"
+    - name: Add sumo ppa
+      run: sudo add-apt-repository ppa:sumo/stable
+    - name: Update apt cache
+      run: sudo apt update
```
SaturnHafen, line 54: We don't need this yet (we are mocking traci), should we remove it?

antonykamp: I would leave it, because it doesn't hurt anybody. And we won't have headaches in the future, because sumo is missing in the pipeline.

src/wrapper/communicator.py
```diff
@@ -48,6 +58,8 @@ def run(self):
 
         traci.start([checkBinary("sumo"), "-c", self._configuration], port=self._port)
 
+        self.sumo_running = True
```
antonykamp, line 61: It was an very good idea with the while loop and this attribute 👍 
Don't forget to set it to `False`, after completing the simulation.

Maybe we should abstract sumo and traci in the naming of variables. For example, we don't name variables connected to the database `postgres_...`. 😇 What do you think about using the terms `simulation_provider` and `simulation_editor` for sumo and traci?

src/wrapper/simulation_component.py
```diff
@@ -0,0 +1,27 @@
+import traci
+
+from src.component import Component
+
+
+class SimulationComponent(Component):
```
antonykamp: `SimulationComponent` is a little bit unspecific in my opinion 😇 What do you think about `ObjectUpdatingComponent` or `SimulationObjectUpdatingComponent`?

src/wrapper/simulation_component.py
```diff
@@ -0,0 +1,27 @@
...
+    Also handles the adding and removing of objects from the simulation.
+    """
+
+    _simulation_objects = None
+
+    def __init__(self, *args, **kwargs):
+        Component.__init__(self, *args, **kwargs)
+        self._simulation_objects = []
+
+        self._fetch_initial_simulation_objects()
```
antonykamp: I have a question 😇 What's the current idea? Where does the information about the initial simulation objects come from? 
I thought we should pass these when initializing the object because the implementor only should have access to the configuration data in the DB (keyword: encapsulation).

antonykamp (commented):
Thank you for your work 👍 
I really like the interface of the new classes. Maybe we need more specific class names, but nothing to worry about.
And I'm not sure if we still need the attribute `sumo_running`, when you don't use it in the tests. Maybe it's something for another PR.

PS: Don't forget a descriptive description of the PR 😄 This helps reviewer allot!

src/wrapper/simulation_component.py
SaturnHafen: The idea is to load them from sumo (see #62 ).

src/wrapper/communicator.py
SaturnHafen: I don't know if that is feasible as the split is probably sumo-specific. E.g. I wouldn't classify Traci as a `simulation_editor`. In any case if you were to implement this interface for another simulator, you would have to rewrite this class anyway, so I'm inclined to leave it as-is.

src/wrapper/simulation_component.py
antonykamp: Thank you for the clarification!

src/wrapper/communicator.py
antonykamp: Understandable 👍

src/wrapper/simulation_component.py
antonykamp: Dont forget to rename the file too 😄

antonykamp (approved):
Looks good to me now 👍 I'm hyped 🙌

PR #63, opened 2023-03-22
Feature: Setup local database

### LINKED ISSUE #61: Setup Local Database
I need to setup a postgres database inside a Dockerimage. Also I have to setup a base class that enables us to use peewee as an ORM and marshmallow for (de)serialization.

## (Wiki) References
- [ORM](https://github.com/BP2022-AP1/bp2022-ap1/wiki/ORM-&-(de)serialization)
- [Database Management](https://github.com/BP2022-AP1/bp2022-ap1/wiki/Database-Management)

## Todos:
- [x] Install dependecies
    - [x] peewee
    - [x] peewee_migrate
    - [x] marshmallow
- [x] Implement base class for ORM and (de)serialization
- [x] setup postgres in Docker

CR1337:
Fixes #61

We now have local databases for dev and test and a script to manage it. Also we have an abstract BaseModel all Models can inherit from.

See [Wiki](https://github.com/BP2022-AP1/bp2022-ap1/wiki/Database-Management)

## PR checklist


CR1337:
The test pipeline fails because there is no database defined yet. This should be done next.

CR1337:
> The test pipeline fails because there is no database defined yet. This should be done next.

Or should this be done inside this PR so that we don't break the pipeline an dev?

src/base_model.py
```diff
@@ -0,0 +1,29 @@
+from peewee import Model, PostgresqlDatabase
+
+
+db: PostgresqlDatabase = PostgresqlDatabase(database="postgres",user="postgres",password="root",host="localhost")
```
antonykamp: We should use environments variables here, because in the future we will have a production database with a better password 👍 An `.env` file is quite common.

db/local/db.py
```diff
@@ -0,0 +1,167 @@
...
+        else:
+            router.run(name)
+            print(f'Ran migration: {name}')
+    elif action == 'rollback':
+        fail_on_name(name)
+        router.rollback(name)
+        print(f'Rolled back migration: {name}')
+
+
+def execute_on_args():
```
antonykamp, line 118: Please add some documentation with available commands and arguments to this file 😇 This would help using it in the future

db/local/db.py
```diff
@@ -0,0 +1,167 @@
...
+
+MIGRATION_DIRECTORY: str = "db/local/migrations"
+
+models: list[Type[BaseModel]] = BaseModel.__subclasses__()
+
+
+def run(args: argparse.Namespace):
+    """Run the database
+
+    :param args: parsed arguments
```
antonykamp: Instead of `parsed arguments` as description, could you please add the content of args (in this case `database_name`). Otherwise this comment wouldn't help much 😇 same with stop, recreate, migration.

src/base_model.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class BaseModel(Model):
```
antonykamp, line 15: I really love the idea of the `BaseModel`! It's a good way to combine marshmellow and peewee quickly! 💚

db/local/db.py
```diff
@@ -0,0 +1,167 @@
...
+    """Drop database tables
+    """
+    db.drop_tables(models)
+    print('Dropped tables')
+
+
+def recreate(args: argparse.Namespace):
+    """Drop and create database tables
+
+    :param args: parsed arguments
```
antonykamp: Could you please add a hint, that `recreate` doesn't need arguments, but has to forward it to `drop` and `create`

src/base_model.py
```diff
@@ -0,0 +1,34 @@
+import marshmallow as marsh
+from peewee import Model, PostgresqlDatabase
+
+db: PostgresqlDatabase = PostgresqlDatabase(
```
antonykamp, line 6: We should use environments variables here, because in the future we will have a production database with a better password 👍 An .env file is quite common.

tests/db/test_db.py
```diff
@@ -0,0 +1,17 @@
...
+from src.base_model import BaseModel, db
+
+
+class ModelTest(BaseModel):
+    """Model for testing purposes"""
+
+    test_value = IntegerField()
+
+
+def test_db():
```
antonykamp: Please add a more descriptive name 😇 e.g. `test_db_connection_workflow`

db/local/db.py
```diff
@@ -0,0 +1,173 @@
...
+from peewee_migrate import Router
+
+from src.base_model import BaseModel, db
+
+MIGRATION_DIRECTORY: str = "db/local/migrations"
+
+models: list[Type[BaseModel]] = BaseModel.__subclasses__()
+
+
+def run(args: argparse.Namespace):
```
antonykamp, line 15: Question: Do we need to validate `args` or does argparse take over the job?

bin/test
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
antonykamp: We could think replacing the bash scripts with 
https://github.com/nat-n/poethepoet

antonykamp (changes requested):
Hey 👋 
I really love the structure and the idea with the `BaseModel` class 💚 Good job!
I just requested some styling and the use of environment variables.
Additionally I proposed some styling changes, nothing to worry about 😄

antonykamp:
> > The test pipeline fails because there is no database defined yet. This should be done next.
>
> Or should this be done inside this PR so that we don't break the pipeline an dev?

I think it's a good idea to add an empty database in this PR 👍

db/local/db.py
antonykamp: Maybe we should flag some arguments as required

antonykamp:
review and merge after https://github.com/BP2022-AP1/bp2022-ap1/pull/74

pyproject.toml
```diff
@@ -22,6 +22,11 @@ pytest = "^7.2.2"
 black = "^23.1.0"
 isort = "^5.12.0"
 pylint = "^2.17.0"
+traci = "^1.16.0"
+peewee = "^3.16.0"
+peewee-migrate = "^1.6.6"
+marshmallow = "^3.19.0"
```
SaturnHafen: Why are these dev-dependencies?

db/local/db.py
```diff
@@ -0,0 +1,173 @@
...
+    """Stop the database
+
+    :param args: parsed arguments containing the string `database_name`
+    :type args: argparse.Namespace
+    """
+    name = args.database_name
+    print("Stopping database...")
+    subprocess.run(
+        [
+            "docker-compose",
```
SaturnHafen: Depending on your docker install, this could also be `docker compose`. `docker-compose` is the older version (See: <https://docs.docker.com/compose/compose-v2/#differences-between-compose-v1-and-compose-v2>)

db/local/db.py
```diff
@@ -0,0 +1,173 @@
...
+
+    :param args: parsed arguments containing the string `database_name`
+    :type args: argparse.Namespace
+    """
+    name = args.database_name
+    print("Running database...")
+    print(os.getcwd())
+    subprocess.run(
+        [
+            "docker-compose",
```
SaturnHafen: Same as below

db/local/db.py
SaturnHafen: This also assumes that you are member of the `docker` group (linux), as otherwise you will not have access to docker without `sudo`

db/local/db.py
```diff
@@ -0,0 +1,173 @@
...
+import subprocess
+from typing import Type
+
+from peewee_migrate import Router
+
+from src.base_model import BaseModel, db
+
+MIGRATION_DIRECTORY: str = "db/local/migrations"
+
+models: list[Type[BaseModel]] = BaseModel.__subclasses__()
```
SaturnHafen, line 12: Is that how it is supposed to work? Seems kind of brittle...

SaturnHafen (commented):
Looks good so far!
We need to document the dependencies on `docker` (and `poetry`, but that might be for a later pr)

pyproject.toml
antonykamp: Good question 👍 We should discuss what a dev dependency defines. A dev dependency doesn't directly impact the product, e.g., linting or command definition. I would add this definition to the wiki if it's fine.

db/local/db.py
antonykamp: Good catch! Should we use v2?

antonykamp: Good catch too! We should add this to the documentation, that you should be part of the docker group.

SaturnHafen: Well v1 will be deprecated end  of june 2023, so yes.

antonykamp: 👍

db/local/db.py
antonykamp: It works 😄 But we can think about declaring an array by hand. This would require changing the array, when declaring a new subclass of `BaseModel`.

SaturnHafen: I guess we can leave it as-is.

antonykamp:
approved? @SaturnHafen

SaturnHafen (approved):
Looks good to me!

antonykamp (approved):

PR #70, opened 2023-03-23
Feature: Setup dotenv

### LINKED ISSUE #67: Setup `dotenv`
## User stories
As a developer I want save my environment variables in a file in order to not share them with other by accident.

## Acceptance Criteria
- [x] Install dotenv-python
- [x] setup constants file
- [x] Update wiki

antonykamp:
Fixes #67 

Install dev dependency `dotenv` to load env variables from files.
Add constants file to test and src folders. These contain environment variables in the future.
Add .env.secret.sample file to define needed env variables.

You can find the edited wiki entry here:
https://github.com/BP2022-AP1/bp2022-ap1/wiki#environment-variables

## PR checklist

- [ ] ~Breaking changes anounced~


src/constants.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+from dotenv import dotenv_values
+
+constants = {
+    **dotenv_values(".env.shared"),  # load shared development variables
+    **dotenv_values(".env.secret"),  # load sensitive variables
+    **os.environ,  # override loaded values with environment variables
+}
```
SaturnHafen: Should we add a test to check if an environment variable appears in the constants dict?

SaturnHafen (approved):
We should probably add an example to the wiki regarding the use of the `constants.py` file. Additionally, I fixed some spelling mistakes in the wiki.

src/constants.py
antonykamp: In my opinion, tests should test the functionality our program, not the development environment. What do you think?

antonykamp:
Thanks for fixing some spelling in the wiki 🙌 I can add an example to the wiki, that's a great idea!

src/constants.py
SaturnHafen: Sounds reasonable. An other issue that might be worth testing for: The preference of loaded env-vars (`os.environ` is more important than `env.secret` is more important than `env.shared`; in case someone restructures this code later).

antonykamp: That sounds reasonable 👍 I'll add it right away

pyproject.toml
```diff
@@ -22,6 +22,7 @@ pytest = "^7.2.2"
 black = "^23.1.0"
 isort = "^5.12.0"
 pylint = "^2.17.0"
+python-dotenv = "^1.0.0"
```
SaturnHafen, line 25: You declared dotenv as dev-dependency, shouldn't it be a normal dependency instead?

antonykamp: It's a dev dependency, because we will use it locally only. It doesn't brings functionality to the user.

SaturnHafen: Fine, but then this import will break in deployment, or am I mistaken?
<https://github.com/BP2022-AP1/bp2022-ap1/pull/70/files#diff-42d0f00c4780305311eae71fecca993a0523587cb4156e039cabd0d1ad8dbbb8R3>

antonykamp: I'm not a poetry-export, but in node dev-dependencies will be installed in production too. That's why I thought, that`poetry install` installs every dependency registered in the `pyproject.toml`. The groups are just a help for the developer.

PR #116, opened 2023-03-28
Feature/91 `ISimulationObjects`

### LINKED ISSUE #91: Add skeleton for `ISimulationObjects` interface
## User stories
As a developer of the Spawner and FaultInjector component I want to know the skeleton of the interface ISimulationObjects in order to use the TraCIWrapper component.

## (Wiki) References
- https://github.com/BP2022-AP1/bp2022-ap1/wiki/Architecture#traciwrapper

## Acceptance Criteria
- [x] Add ISimulationObjects class as skeleton 
- [x] Update wiki

SaturnHafen:
Fixes #91 

## PR checklist

- [ ] ~~Additional features are tested~~
- [ ] ~~Breaking changes anounced~~


src/interfaces.py
```diff
@@ -1,5 +1,105 @@
...
+
+    def add_subscriptions(self) -> int:
+        """Gets called when this object is created to allow
+        specification of simulator-synchronized properties.
+
+        :return: The synchronized properties (see <https://sumo.dlr.de/pydoc/traci.constants.html>)
+        """
+
+
+class ISimulationNode(ISimulationObject, Protocol):
```
antonykamp, line 25: Doesn't get `Protocol` inherited via `ISimulationObject` to `ISimulationNode`(and the other classes) too? 🤔

src/interfaces.py
```diff
@@ -1,5 +1,105 @@
...
+    """A point somewhere in the simulation where `ISImulationTrack`s meet"""
+
+    position: Tuple[float, float]
```
antonykamp, line 28: Does users of the interface have access to the properties? 🤔 Otherwise it should be part of the interface, I think, because the user shouldn't know the concrete functionality. 😇 But if so, then please ad the functions with the `@property`-tag
Sorry for the spongy definition of the term "interface" 😞

src/interfaces.py
```diff
@@ -1,5 +1,105 @@
...
+    current_track: ISimulationTrack
+    priority: int
+    position: Tuple[float, float]
+    speed: float
+    max_speed: float
+    route: str
+    timetable: List[ISimulationPlatform]
+    _train_type: str
+
+    def __init__(
```
antonykamp, line 87: Why does `ISimulationTrain` have a `__init__`function and the other Protocols doesn't? 🤔 Even the `ISimulationObject`...

antonykamp (commented):
The interface is very clean 👍 Great! 
I just have some questions regarding properties, the inheritance of `Protocol` and an `__init__` function. But I have to say, that I'm not a python expert...soo....yes 😄 Thanks for the change!

src/interfaces.py
antonykamp: @CR1337 `__init__` should be part of a Protocol, shouldn't it?

src/interfaces.py
antonykamp: @CR1337 What does our Protocol-expert says about properties in Protocols? 🤔

src/interfaces.py
SaturnHafen: It is only here because this is the only method that has a implementor-facing creation option (we could implement a class-method `spawn_train` or something instead). In every other object, the `__init__` is part of the private api.

src/interfaces.py
SaturnHafen: I am not sure. I read [this post](https://stackoverflow.com/questions/69783371/python-multiple-inheritance-from-class-and-protocol), but re-reading it does not state that one would need to use multi-inheritance ...

src/interfaces.py
SaturnHafen: See <https://peps.python.org/pep-0544/#protocol-members>
This might have to be changed anyway, as some properties should be read-only.

src/interfaces.py
CR1337: When you inherit from a Protocol the resulting subclass won't be a Protocol anymore. Therefore if you want to have it to be a Protocol you have to explicitly inherit from it again.

See https://peps.python.org/pep-0544/#merging-and-extending-protocols

src/interfaces.py
CR1337: > Does users of the interface have access to the properties? thinking Otherwise it should be part of the interface, I think, because the user shouldn't know the concrete functionality. innocent But if so, then please ad the functions with the `@property`-tag Sorry for the spongy definition of the term "interface" disappointed

In my opinion any element (attribute, method, property, ...) should be part of the protocol when an outside object should be able to access it.

CR1337: > See https://peps.python.org/pep-0544/#protocol-members This might have to be changed anyway, as some properties should be read-only.

Properties are always readonly as long as you only specify a `@property`-decorator. You have to make it writable explicitly. For example for a property `test` you would have to define another method decorated with `@test.setter`.

src/interfaces.py
CR1337: Isn't `__init__` always public? 
Yes I could think of situations where it would only be called from inside the class itself but normally it is called from the outside.

PEP 544 doesn't seem to state it explicitly but the examples suggest that `__init__` should not be part of a protocol.

CR1337: On the other hand it is important information how `__init__` works. I would say practicality over dogma and lets have the  `__init__` inside the Protocol when construction is part of the interface?

src/interfaces.py
antonykamp: Good to know 👍 Thanks for the clarification! I learn something new every day.

src/interfaces.py
antonykamp: Ahhhh...gracias! 💚

src/interfaces.py
antonykamp: Sounds reasonable 🤔 Then we have to edit some existing Protocols, correct?

antonykamp:
Do we want to close this PR, because we don't use Protocols anymore?

PR #120, opened 2023-03-29
Feature/87 i spawner

### LINKED ISSUE #87: Add skeleton of `ISpawner` interface
## User stories
As a developer of the TraCICommunicator component I want to know the skeleton of the interface ISpawner in order to use the Spawner component.

## (Wiki) References
-https://github.com/BP2022-AP1/bp2022-ap1/wiki/Architecture#spawner

## Acceptance Criteria
- [x] Add ISpawner
- [x] Update wiki

CR1337:
Fixes #87 

## PR checklist


CR1337:
The linter fails because it needs to be configured ignoring docstrings in the interfaces.

src/interfaces.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any, Protocol
+
+
+class IComponent(Protocol):
+    logger: "ILogger"
```
instant-sky, line 6: When exactly do you use "" in TypeHints?

instant-sky (approved):
Looks good, very clean interface!

src/interfaces.py
CR1337: I can think of two situations where string type hints are used.

1. When the type doesn't exist yet where the hint is used. That is the case for example when a class has a reference to itself:

```python
class Foo:
    other_foo: "Foo"
```

Because `Foo` doesn't exist yet during its own definition you have to use a string for type hinting.

2. When the type isn't defined yet.

Here the reason is the second case. `ILogger` doesn't exist yet. Therefore I use a string. When `ILogger` is defined I will replace the string with the real type.

antonykamp:
Do we want to close this PR, because we don't use Protocols anymore?

PR #148, opened 2023-03-31
Refactor: Use BaseModel from example

### LINKED ISSUE #146: Refactor: Use BaseModel from example
## User stories
As a developer I want to work with a handy BaseModel in order to implement new feature quickly.
Therefore, please take a look into the example.

## (Wiki) References
- https://github.com/BP2022-AP1/bp2022-ap1/wiki/ORM-&-(de)serialization
- https://github.com/CR1337/flask-peewee-marshmallow-demo/blob/main/src/model/base_model.py

## Acceptance Criteria
- [x] Replace current BaseModel with new one
- [x] Add test for serialization
- [x] Add test for deserialization
- [x] Add test for DB change

antonykamp:
Fixes #146 

Add abstraction of serialization and deserialization to BaseModel.
Added tests and introduces TestDB class in order to use setup and teardown methods.

## PR checklist

### DIFF
--- a/src/base_model.py
+++ b/src/base_model.py
@@ -1,7 +1,9 @@
+import abc
 import os
+from uuid import uuid4
 
 import marshmallow as marsh
-from peewee import Model, PostgresqlDatabase
+from peewee import Model, PostgresqlDatabase, SqliteDatabase, UUIDField
 
 db: PostgresqlDatabase = PostgresqlDatabase(
     database=os.getenv("DATABASE_NAME") or "postgres",
@@ -18,10 +20,24 @@ class BaseModel(Model):
     class Meta:
         """Set Database"""
 
-        database = db
+        database: SqliteDatabase = db
 
     class Schema(marsh.Schema):
         """The marshmallow schema all model schemas have to inherit from."""
+
+        id = marsh.fields.UUID()
+
+        @abc.abstractmethod
+        def _make(self, data: dict) -> "BaseModel":
+            """Constructs a model object from a dictionary."""
+            raise NotImplementedError()
+
+        @marsh.post_load
+        def make(self, data: dict, **_) -> "BaseModel":
+            """Constructs a model object from a dictionary."""
+            return self._make(data)
+
+    id = UUIDField(primary_key=True, default=uuid4)
 
     @classmethod
     def from_dict(cls, data: dict) -> "BaseModel":
--- a/tests/db/test_db.py
+++ b/tests/db/test_db.py
@@ -1,3 +1,6 @@
+from uuid import UUID
+
+import marshmallow as marsh
 from peewee import IntegerField
 
 from src.base_model import BaseModel, db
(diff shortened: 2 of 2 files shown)


antonykamp:
Please don't forget to reference @CR1337 as co author (like we did 💚)

tests/db/test_db.py
```diff
@@ -6,12 +9,43 @@
...
+        db.create_tables([ModelTest])
+
+    def teardown_method(self):
+        db.drop_tables([ModelTest])
+
+    def test_db_connection_workflow(self):
+        # I just wanted to test if the db is working
+        ModelTest.create(test_value=1).save()
+        test_obj = ModelTest.select().where(ModelTest.test_value == 1).first()
+        assert test_obj.test_value == 1
```
Lietze, line 40:
```suggestion
        test_obj = ModelTest.create(test_value=1)
        test_obj.save()
        returned_obj = ModelTest.select().where(ModelTest.test_value == 1).first()
        assert test_obj == returned_obj
```
It may be better to check for the identity of the returned object and not fot the value.

tests/db/test_db.py
```diff
@@ -6,12 +9,43 @@
...
-    # I just wanted to test if the db is working
-    db.create_tables([ModelTest])
-    ModelTest.create(test_value=1).save()
-    test_obj = ModelTest.select().where(ModelTest.test_value == 1).first()
-    assert test_obj.test_value == 1
+    def test_deserialization(self):
+        # I just wanted to test if the deserialization is working
+        test_obj = ModelTest.Schema().load({"test_value": 1})
+        assert test_obj.test_value == 1
+        assert isinstance(test_obj.id, UUID)
```
Lietze, line 51: Why do you only check if it is a instance and not check if a specific UUID can be loaded?

src/base_model.py
```diff
@@ -18,11 +20,25 @@ class BaseModel(Model):
...
 
     class Schema(marsh.Schema):
         """The marshmallow schema all model schemas have to inherit from."""
 
+        id = marsh.fields.UUID()
+
+        @abc.abstractmethod
+        def _make(self, data: dict) -> "BaseModel":
+            """Constructs a model object from a dictionary."""
+            raise NotImplementedError()
```
Lietze, line 33: Should this method load the UUID? I think, that if that is not the case, the UUIDs will not be stable when reloading the objects.

Lietze (changes requested):
Nice job, this looks like a good addition.

src/base_model.py
Lietze: Or is that already done by a subclass method calling `return ModelTest(**data)`?
If so, could a method like that already be defined in the BaseModel?

antonykamp: We shouldn't be able to define the id when initializing an object from a dict because we could overwrite existing rows in the database. When you want to update an object, run a query and update it 😅

tests/db/test_db.py
antonykamp: I would keep comparing each value, because it's not the same object. Querying a ModelTest instance creates a new object with the same value.

tests/db/test_db.py
antonykamp: Because we don't have any control about the value of the uuid. It's generated by peewee after loading the dict. 😇

Lietze (approved):
Thanks for the clarifiction.

PR #151, opened 2023-03-31
Fix: replace __subclasses__ class with constant

### LINKED ISSUE #149: Doesn't create tables on `poe db create`
## Problem Description

1. What happens
2. What is the expected behaviour
3. If not VERY clear: Why is this a problem

When running `poe db create` no tables where initialized, becuase `BaseModel.__subclasses__()` only load already loaded classes. Therefore we either have to import every class that needs its own table or introduce a list of needed tables in a constants file (preferred way). 

## Steps For Reproduction

1. Add a class that inherits from `BaseModel`
2. run `poe db create`
3. Look into tables, e.g., with pgAdmin

antonykamp:
Fixes #149 

Replace __subclasses__ call with constant list containing needed tables.
Moves into own file to prevent circular imports.

## PR checklist

### DIFF
--- a/db/local/db.py
+++ b/db/local/db.py
@@ -1,15 +1,13 @@
 import argparse
 import os
 import subprocess
-from typing import Type
 
 from peewee_migrate import Router
 
-from src.base_model import BaseModel, db
+from src.base_model import db
+from src.constants import tables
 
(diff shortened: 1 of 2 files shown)


instant-sky (approved):
looks good!

PR #192, opened 2023-04-04
Feature/98 add basic fahrstrssen workflow

### LINKED ISSUE #98: Add basic workflow for setting new fahrstrassen for trains continuously
## User stories
As a user, I want to have a basic `RouteController` workflow, that sets new Fahrstrassen when a train is close to the end of his current Fahrstrasse in order for one or a few hard-coded examples to work.

## Dependencies
Depends on:
 - #97 
 - #92 
 - #89 #90 #95 #96
 - #26

## (Wiki) References
- https://github.com/BP2022-AP1/bp2022-ap1/wiki/Architecture#interlocking

## Acceptance Criteria
- [x] it is triggered by a train entering a new track_segment
- [x] then it is checked if the train is close to the end of his Fahrstrasse (by checking if the current segment is the last of the active route)
- [x] if so, a new Fahrstrasse is set for the train
- [x] along a Route calculated using dijkstra
- [x] and the old Fahrstrasse is freed

Lietze:
Fixes #98 

## Description
I added a happy case workflow for setting a fahrstrasse. This includes djikstra to find the route. If you are confused by the diffrent ids of objects from different models, you may want to take a look at [this](https://github.com/BP2022-AP1/bp2022-ap1/wiki/Interlocking-Component#relationships-between-data-models) diagramm I made.

## PR checklist


src/interlocking_component/route_controller.py
```diff
@@ -106,16 +109,53 @@ class RouteController(IRouteController):
...
         """
-        raise NotImplementedError()
+        route = None
+        for route_candidate in self.interlocking.active_routes:
+            interlocking_track_candidat = route_candidate.contains_segment(
+                track.identifier
+            )
+            # This assumes, that track.identifier does not contain '-re',
+            # as the interlocking does not know of reverse directions.
+            if interlocking_track_candidat is not None:
```
antonykamp, line 132: Hypothetical question: Can multiple routes contain the segment in question?

src/interlocking_component/route_controller.py
```diff
@@ -106,16 +109,53 @@ class RouteController(IRouteController):
...
+                route = route_candidate
+        if route is None or route.get_last_segment_of_route != track.identifier:
+            return False
+
+        new_route = self.router.get_route(track, train.timetable[0].track)
```
antonykamp, line 137: Please correct me, if I'm wrong :D 
Isn't  `train.timetable[0]` the initial location of the train? 🤔 
That would mean that `new_route` is a list of nodes with a path from the current track to the starting point.

src/interlocking_component/route_controller.py
```diff
@@ -106,16 +109,53 @@ class RouteController(IRouteController):
...
+        for end_node_candidat in new_route:
+            for interlocking_route in self.interlocking.routes:
+                if (
+                    interlocking_route.start_signal.name == new_route[0]
```
antonykamp, line 143: I love your comments 💚 That makes is easy to understand the code!
Could you please add a small comment to this statement too? 🙃

src/interlocking_component/route_controller.py
```diff
@@ -106,16 +109,53 @@ class RouteController(IRouteController):
     It makes sure, that the Interlocking sets fahrstrassen along those routes.
     """
 
-    def is_new_fahrstrasse_needed(self, train: "Train", track_segment: "Track"):
+    interlocking: interlockinginterface = None
+    router: Router = None
+
+    def is_new_fahrstrasse_is_needed(self, train: Train, track: Track) -> bool:
```
antonykamp: You have to `is` in the name of this function. Do you need it? 🙃

src/interlocking_component/router.py
```diff
@@ -1,12 +1,51 @@
...
+        penultimate_node = end_track.from_node
+        # The next part is dijkstra as a first mvp.
+        # This will lead to deadlocks if two trains drive in opposite directions.
+        distances = {}
+        previous_nodes = {}
+        distances[start_node] = 0
+        current_index = 0
+        current_node = start_node
+        while True:
+            sorted_distances = sorted(distances.items(), key=lambda item: item[1])
```
antonykamp, line 33: What does `item[1]` contain? Could you please add a small comment about the sorting or replace the `1` with an semantic constant 👍

src/interlocking_component/router.py
```diff
@@ -1,12 +1,51 @@
...
+            last_node = current_node
+            current_node = sorted_distances[current_index][0]
```
antonykamp, line 38: What does `current_node = sorted_distances[current_index][0]` contain? Could you please add a small comment about the sorting or replace the`0` with a semantic constant 👍

poetry.lock
antonykamp, line 1: We'll leave it as it is :D

antonykamp (commented):
I have some questions 😇 But without much knowledge about the Interlocking component, I could read your code flawlessly 🙃 
Is there a reason why you didn't add tests to your code 😄 It's quite an important part, IMO.

src/interlocking_component/route_controller.py
Lietze: No, I think you are right, that is not needed. This is the method, that checks and will handle if the fahrstrasse could not be set, so there is no reason to return if it was successful.

src/interlocking_component/route_controller.py
Lietze: Multiple routes contain the same segment, but only one of those routes at a time can be in active_routes. So this is not a problem.

src/interlocking_component/route_controller.py
Lietze: This was not defined before. In the future the next plattform the train should drive to will always be at ```train.timetable[0]``` and when the train arives at a plattform that one will be deleted from the list.
This is the new issue: #314

src/interlocking_component/router.py
Lietze: ```distances``` is a dict, so each ```item``` is one entry in the dict. ```item[1]``` then referes to the value of the dict entry. So this sorts the dict based on the values. I will add a comment here.

src/interlocking_component/router.py
Lietze: I hope the comment I added above also helped with this.

src/interlocking_component/router.py
antonykamp: But what is `item[0]`? :D A route? A number? 🙃

Lietze:
> I have some questions 😇 But without much knowledge about the Interlocking component, I could read your code flawlessly 🙃 Is there a reason why you didn't add tests to your code 😄 It's quite an important part, IMO.

Hey @antonykamp, thank you for the review. Yes there is a reason why I didn't add test, although it may be better to add them. In order to test ```maybe_update_fahrstrasse```, I either need to mock the interlocking or implement the initialisation of the interlocking first.

To mocking the interlocking is possible, but I would rather test if the systems can interact with each other. Also, I don't think there is much to test other than the interaction.

Another big problem for testing this right now is, that the graph-like properties of the nodes and tracks in the traci-wrapper are not yet implementet. This means, that I can not test the router right now.

If this gets merged without the tests, I will add issues for this.

antonykamp:
Thanks for sharing your thought! Sounds really well thought out 👍 

> If this gets merged without the tests, I will add issues for this.

I think that's the best solution.

antonykamp (approved):

PR #217, opened 2023-04-05
Log entry schema

### LINKED ISSUE #14: Log Entry Schema
## User stories
As a developer I want to store logging information in a database by having a defined Schema

## Dependencies
Depends on: 
 - #7 

## (Wiki) References
- TBD

## Acceptance Criteria
- [ ] Logger Entry should be clearly defined in wiki
- [ ] Logger Entry should be implemented as Class in ORM

LucasDerReisende:
Fixes #14 

## PR checklist


src/logger/log_entry.py
```diff
@@ -0,0 +1,166 @@
...
+        message = marsh.fields.String(required=True)
+        run_id = marsh.fields.UUID(required=True)
+
+        def _make(self, data: dict) -> "LogEntry":
+            return LogEntry(**data)
+
+    timestamp = DateTimeField(null=False, default=datetime.now())
+
+    message = CharField(null=False)
+    run_id = ForeignKeyField(Run, null=False)
```
antonykamp, line 43: If you think it's possible that a run needs a reference to its log entries, please add a `backRef` 😇 
https://docs.peewee-orm.com/en/latest/peewee/relationships.html

src/logger/log_entry.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class TrainSpawnLogEntry(LogEntry):
+    """A LogEntry that represents the spawning of a train."""
+
+    class Schema(LogEntry.Schema):
+        """The marshmallow schema for the TrainSpawnLogEntry model."""
+
+        train_id = marsh.fields.Integer(required=True)
```
antonykamp: Please discuss with @SaturnHafen or @instant-sky if the id auf a train is an integer or uuid 😇 I'm unsure at the moment.
(Please update it in the other classes too 🥳 )

src/logger/log_entry.py
```diff
@@ -0,0 +1,166 @@
...
+        """The marshmallow schema for the TrainArrivalLogEntry model."""
+
+        train_id = marsh.fields.Integer(required=True)
+        station_id = marsh.fields.Integer(required=True)
+
+        def _make(self, data: dict) -> "TrainArrivalLogEntry":
+            return TrainArrivalLogEntry(**data)
+
+    train_id = IntegerField(null=False)
+    station_id = IntegerField(null=False)
```
antonykamp: Please discuss with @SaturnHafen or @CR1337 if the id of a platform/station is an integer, uuid or string 😇 I'm unsure at the moment.
(Please update it in the other classes too 🥳 )

src/logger/log_entry.py
antonykamp: Please don't forget how "Abstract" classes will be represented as a table in the DB. Peewee will create for each class (abstract and subclasses) their own table. That means back referencing wouldn't work. If it's not needed, it's not a problem.

src/logger/log_entry.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class CreateFahrstrasseLogEntry(LogEntry):
+    """A LogEntry that represents the creation of a fahrstrasse."""
+
+    class Schema(LogEntry.Schema):
+        """The marshmallow schema for the CreateFahrstrasseLogEntry model."""
+
+        fahrstrasse = marsh.fields.String(required=True)
```
antonykamp, line 112: Are marshmellow String and peewee-charfield compatible? 😇 I'm unsure 🤔

src/logger/log_entry.py
```diff
@@ -0,0 +1,166 @@
...
+    fahrstrasse = CharField(null=False)
+
+
+class SetSignalLogEntry(LogEntry):
+    """A LogEntry that represents the setting of a signal."""
+
+    class Schema(LogEntry.Schema):
+        """The marshmallow schema for the SetSignalLogEntry model."""
+
+        signal_id = marsh.fields.String(required=True)
```
antonykamp: Please discuss with @SaturnHafen or @Lietze  if the id of a signal is an integer, uuid or string 😇 I'm unsure at the moment.
(Please update it in the other classes too 🥳)

src/logger/log_entry.py
```diff
@@ -0,0 +1,166 @@
...
+        injection_id = marsh.fields.Integer(required=True)
+        fault_type = marsh.fields.Integer(required=True)
+        injection_position = marsh.fields.String(required=True)
+        duration = marsh.fields.Integer(required=True)
+
+        def _make(self, data: dict) -> "InjectFaultLogEntry":
+            return InjectFaultLogEntry(**data)
+
+    injection_id = IntegerField(null=False)
+    fault_type = IntegerField(null=False)
```
antonykamp: Please discuss with @instant-sky how he identifies a faulttype and injection. Is it an integer, uuid or string 😇 I'm unsure at the moment.
(Please update it in the other classes too 🥳 )

antonykamp (commented):
The separation in classes an the tests are great 🥳 I'm reading good and hard work here 👍 
Please discuss the types of train_id, station_id (probably platform_id), faut_type, injection_id with the stakeholder :)

src/logger/log_entry.py
SaturnHafen: The id is probably neither. Please discuss with @CR1337 for further details

src/logger/log_entry.py
SaturnHafen: `yaramo` uses `uuid`s so we probably will stick with that (I am currently unsure if these are shorted)

src/logger/log_entry.py
SaturnHafen: As these are not generated by yaramo, we should probably stick with human-readable identifiers (string-names)

src/logger/log_entry.py
LucasDerReisende: I think, I do not want to use references for now. If I need them later, I will add them to all classes.

src/logger/log_entry.py
LucasDerReisende: Will change it to a string :)

src/logger/log_entry.py
LucasDerReisende: Will also change it to a string :)

src/logger/log_entry.py
LucasDerReisende: Will change CharField to TextField since they allow variable length

src/logger/log_entry.py
LucasDerReisende: Will change it to UUID

src/logger/log_entry.py
LucasDerReisende: Is discussed, will be changed

src/logger/log_entry.py
```diff
@@ -120,14 +130,14 @@ class SetSignalLogEntry(LogEntry):
...
-        signal_id = marsh.fields.String(required=True)
+        signal_id = marsh.fields.UUID(required=True)
         state_before = marsh.fields.Integer(required=True)
         state_after = marsh.fields.Integer(required=True)
 
         def _make(self, data: dict) -> "SetSignalLogEntry":
             return SetSignalLogEntry(**data)
 
-    signal_id = CharField(null=False)
+    signal_id = TextField(null=False)
```
antonykamp: You can use `UUIDField` here 😇

src/logger/log_entry.py
```diff
@@ -14,15 +23,16 @@ class Schema(BaseModel.Schema):
...
         message = marsh.fields.String(required=True)
         run_id = marsh.fields.UUID(required=True)
 
         def _make(self, data: dict) -> "LogEntry":
             return LogEntry(**data)
 
     timestamp = DateTimeField(null=False, default=datetime.now())
-
-    message = CharField(null=False)
+    tick = IntegerField(null=False)
```
antonykamp: Do you think integer is enough? 🤔 To get meaningful data, we need to run the simulation for some time.

antonykamp (commented):
Looks like quite some changes 🙌 Great! I'm really happy to have you back! 
Please take a look at two peewee field 😇 

(Great easter and allot of good food  wishes you the little doggo addicted girl - charly ;))

src/logger/log_entry.py
LucasDerReisende: I don't know what you mean exactly. In python integers are unbound and in PostgreSQL e.g. the limit is at 2147483647, which should be plenty to run some simulations if we assumed that every tick more or less represents a second in the simulation.

src/logger/log_entry.py
LucasDerReisende: Did not know, it existed, good to know :) I'll change it

antonykamp: Perfect 🥳

src/logger/log_entry.py
antonykamp: Sorry that I only wrote such a short description 😄 
I meant that we must have long simulation runs to collect meaningful data. That means the ticks shouldn't limit us. If somebody wants to simulate over 70 years, then the person shouldn't be restricted. As developers, it's not a big deal to use BigInteger Fields in the database. But for the user, it could be a pain.
That's my opinion :D I understand the point that nobody would simulate over 70 years 😂 
You decide 😇 It was just an idea.

LucasDerReisende: I mean, it is no problem to change it, it might be good in the future to have this possibility

LucasDerReisende: Should be changed now, I'm assuming that the Marshmallow integer field corresponds with the Python integer field, in that they both allow for unbound values

tests/logger/test_log_entry.py
```diff
@@ -326,6 +332,9 @@ def test_serialization(self, train_remove_log_entry_as_dict):
 
             assert train_remove_log_entry.to_dict() == {
                 "id": str(train_remove_log_entry.id),
+                # pylint: disable=no-member
```
antonykamp, line 335: Just from curiosity, why do you have to disable the linter? 🤔

LucasDerReisende: We are using DateTimeField which corresponds with a DateTime object, but the linter does not recognize this. I think we had the same problem last week. If you have an idea how to solve this, I am open, but I did not find anything.

antonykamp: Thanks! 💚

antonykamp (approved):

PR #235, opened 2023-04-06
Feature/225 created updated base model

### LINKED ISSUE #225: Created and Updated in BaseModel
## User stories
As a developer I want to have access to `created_at` and `updated_at` fields in order to performs certain operations with it e.g. sorting by date.

## Acceptance Criteria
- [ ] implement `created_at` field
- [ ] implement `updated_at` field
- [ ] override `save` in `BaseModel` to automatically update `updated_at`

CR1337:
Fixes #225

`BaseModel` now has two new fields: `created_at` and `updated_at`. `created_at` is automatically set at object creation. `updated_at` is set on every call of `save` or `create`.

See [Wiki](https://github.com/BP2022-AP1/bp2022-ap1/wiki/ORM-&-(de)serialization#basemodel)

## PR checklist


src/base_model.py
```diff
@@ -26,6 +27,8 @@ class Schema(marsh.Schema):
         """The marshmallow schema all model schemas have to inherit from."""
 
         id = marsh.fields.UUID()
+        created_at = marsh.fields.DateTime(format='iso')
```
antonykamp: Discussion: Should somebody be able to set updated_at and created_at? 🤔

antonykamp (commented):
Looks really great 🙌 Hyypppeeee
I just have one question/discussion regarding the accessibility of the new fields from marshmellow 😇

src/base_model.py
CR1337: Readonly access to `updated_at` and `created_at` during (de)serialization would be a good idea. Unfortunately marshmallow doesn't support this feature natively. But there are workarounds. Let me try something. I will be back 😎

antonykamp: Thanks for the insights 👍 I'm looking forward to your little experiment ;)

CR1337: Update: There is functionality for this but it will just ignore the fields when given in a dict. 

Do we want an exception to be raised in this case? If yes, we need a custom construct (which I already have an idea for).

antonykamp: > Do we want an exception to be raised in this case? If yes, we need a custom construct (which I already have an idea for).

I'm personally not a fan of over-configuration of a tool 😇 

> Update: There is functionality for this but it will just ignore the fields when given in a dict.

But I'm a little fan of this 😄 Can you send me a link to the docs please 😇 I'm really interested

CR1337: Here are the docs for readonly and writeonly fields for marshmallow:
https://marshmallow.readthedocs.io/en/stable/quickstart.html#read-only-and-write-only-fields
It is not that much documentation to be honest.

This post deals with a different topic but uses this functionality and states that fields with `dump_only=True` are just ignored during deserialization:
https://stackoverflow.com/questions/70328646/equivalent-of-marshmallow-dump-only-fields-for-pydantic-fastapi-without-multiple

(Also fields with `load_only=True` will be ignored during serialization.)

This post suggests a simple solution for throwing an exception if the dict should contain a readonly field: https://github.com/marshmallow-code/marshmallow/issues/61#issuecomment-63582625

What would be the better solution in your oppinion?

antonykamp: Thank you very much for the research! 

In my opinion, we should use `dump_only=True` for `created_at` and `updated_at`. This is even used as an example 😄 
Screenshot 2023-04-07 at 12 39 09

I'm unsure if we should throw an exception additionally. The case, that somebody tries to define created_at and updated_at by their own, is unlikely. Two safety mechanisms are a bit to much for, aren't they? And I like the simplicity of `dump_only` 😇 

What do you think?

CR1337: I would only have implemented one of the mechanisms. Two safety mechanisms would really be too much. But as `dump_only` is much simpler I tend to use it too.

antonykamp: Great 👍 What do you think about adding to to `id` too? Nobody should be able to define or update the id.

tests/test_table_basic.py
antonykamp, line 1: You can remove it now, because we don't use it anymore 😇

src/base_model.py
CR1337: Sounds good. I will add it.

Update: Well during testing it came out that the stackoverflow post was wrong. When deserialization gets a dump_only-field it raises a `peewee.ValidationError`.

antonykamp: > Update: Well during testing it came out that the stackoverflow post was wrong. When deserialization gets a dump_only-field it raises a peewee.ValidationError.

Thanks for the information! 😇

tests/test_table_basic.py
CR1337: Removed 🙂

src/base_model.py
CR1337: Have a look again :)

antonykamp: Will accept it after merging dev into this branch :P

CR1337:
@antonykamp 
I think I finally did it 😄

antonykamp:
The coverage decreased a bit 😇 Are you able to fix this? If the reason are more `NotImplementedErrors`, then I would approve it 😂

CR1337:
@antonykamp 
It's because I deleted `test_table_basic.py` I guess 🤷😄

antonykamp:
naaaahhhh ;)

Screenshot 2023-04-07 at 13 54 44

CR1337:
> naaaahhhh ;)
> 
> Screenshot 2023-04-07 at 13 54 44

Now all checks passed 😉

antonykamp (approved):
But I'm happy 😇 If you want to take a deeper look into the deceased coverage, I wouldn't prevent you ;)

antonykamp:
You rock 🚀

CR1337:
> But I'm happy innocent If you want to take a deeper look into the deceased coverage, I wouldn't prevent you ;)

I already planned to dig deeper into the test coverage of my code. As other files are also affected I would do it in a separate issue :)

PR #241, opened 2023-04-08
Feature: Add foreign keys between `SimulationConfiguration`, `Run` and `Token`

### LINKED ISSUE #224: Add forgein keys between SimulationConfiguration, Run, Token
## User stories
As a developer I want to have the SimulationConfiguration, Run and Token to be connected in order to reflect every information in the database

## Dependencies
 - #139 
 - #138 
 - #137 
 - #227 

## (Wiki) References
- https://github.com/BP2022-AP1/bp2022-ap1/wiki/Architecture#database-class-diagram

## Acceptance Criteria
- [x] Add foreign key and back reference from Run to SimulationConfiguration
- [x] Add foreign key and back reference from SimulationConfiguration to Token
- [x] Add tests

antonykamp:
Fixes #224 

Add foreign keys from `Run` to `SimulationConfiguration` and `SimulationConfiguration` to `Token`.

## PR checklist

### DIFF
--- a/tests/implementor/test_table_run.py
+++ b/tests/implementor/test_table_run.py
@@ -4,14 +4,10 @@ import marshmallow as marsh
 import peewee
 import pytest
 
-from src.implementor.models import Run
+from src.implementor.models import Run, SimulationConfiguration, Token
 from tests.decorators import recreate_db_setup
 
 
-@pytest.mark.parametrize(
-    "init_dict",
-    [],
-)
 class TestRunFailingInit:
     """Test that a object of a class cannot be created or deserialized with invalid data."""
 
@@ -19,11 +15,26 @@ class TestRunFailingInit:
     def setup_method(self):
         pass
 
+    @pytest.mark.parametrize(
+        "init_dict",
+        [
+            {  # simulation_configuration does not exists
+                "simulation_configuration": "00000000-0000-0000-0000-000000000000",
+            },
+            {},  # simulation_configuration is missing
+        ],
+    )
     def test_create(self, init_dict: dict):
         """Test that an object of a class cannot be saved."""
         with pytest.raises(peewee.IntegrityError):
             Run(**init_dict).save(force_insert=True)
 
+    @pytest.mark.parametrize(
+        "init_dict",
+        [
+            {},  # simulation_configuration is missing
+        ],
+    )
     def test_deserialization(self, init_dict: dict):
         """Test that an object of a class cannot be deserialized."""
         with pytest.raises(marsh.exceptions.ValidationError):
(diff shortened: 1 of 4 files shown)


tests/implementor/test_table_run.py
```diff
@@ -37,20 +48,37 @@ class TestRunSuccessfulInit:
     def setup_method(self):
         pass
 
+    @pytest.fixture
+    def token(self):
+        token = Token(name="owner", permission="admin", hashedToken="hash")
+        token.save()
```
LucasDerReisende, line 54: does there need to be a `.save()`, just curious which you added it

LucasDerReisende (commented):
looks good, I'm just interest, why you used `.save()` in the tests

tests/implementor/test_table_run.py
antonykamp: You can insert a new entry to the database by using `Token.create()` or by initializing it first and saving it with `.save()` later 😇

LucasDerReisende: Yes, but does it need to be saved in the database for this test? Just asking, because I did not in my tests and it also worked

antonykamp: I thought it could cause an error, when I want to reference a Token from a SimulationConfiguration with foreign key (see the following test case) 

https://github.com/BP2022-AP1/bp2022-ap1/blob/6263c730a63dce9038242b96eb88eba1b9567441/tests/implementor/test_table_run.py#L21)

LucasDerReisende: Nevermind, I though about something else, but does not seem to be a problem, so I'm good :)

LucasDerReisende (approved):
Looks good 🔥  keep up the good work

tests/implementor/test_table_run.py
antonykamp: Great 🙌 Thanks for the question, tho 😄

tests/implementor/test_table_run.py
```diff
@@ -37,20 +48,37 @@ class TestRunSuccessfulInit:
     def setup_method(self):
         pass
 
+    @pytest.fixture
+    def token(self):
+        token = Token(name="owner", permission="admin", hashedToken="hash")
```
SaturnHafen, line 53: Should we add validation to prevent not hashed tokens from being entered in the database?

SaturnHafen (commented):
A little later but: Why do we need this "join"?
Is it to create an audit-trail to trace which token created which simulation run/configuration?

antonykamp:
> A little later but 

It's never too late ;) Thanks for the hints! 🙃 

>  Why do we need this "join"? Is it to create an audit-trail to trace which token created which simulation run/configuration?

Exactly! Otherwise, we couldn't tell which simulation configuration belongs to which user or which run belongs to which simulation configuration.

tests/implementor/test_table_run.py
antonykamp: Is it possible to distinguish between a random string and a hash other than prepending `hash`? 🤔 

IMO it's not needed because we insert the tokens by ourselves. 🙃 We'll generate the token, save a hashed version in the database and return the original to the user.

SaturnHafen: Well if you look at (for example) a sha256 hash, it will look consistent (64 hexadecimal chars). If one were to add salt & pepper to the hash, the format would get even more obvious.
In any case, setting a parameter called `hashedToken` to something obviously not hashed, seems a little off to me :slightly_smiling_face:

PR #256, opened 2023-04-11
Feature/215 cleanup trains

### LINKED ISSUE #215: Cleanup trains
## User stories
As a developer I want a better train-API in order to have a consistent api

## Dependencies
Depends on: 
 - #26 
 - #28 

## Acceptance Criteria
- [x] `SimulationObjectUpdatingComponent` needs a property on the trains
  - [x] tests
- [x] The `track` property should return a `Track`, not a string, and should get updated regularly (at least on every access)
  - [x] tests

SaturnHafen:
Fixes #215 

Updates the train to return domain objects

## PR checklist


src/wrapper/simulation_objects.py
```diff
@@ -298,6 +299,12 @@ def track(self) -> Track:
 
         :return: The current track the train is on
         """
+        if self._track is None or self._track.identifier != self._track_id:
```
CR1337, line 302: Is it possible that there is no `item` in `self.updater.tracks` with `item.identifier == self._track_id`?
If yes, what happens then?

SaturnHafen: Then this method will crash (which will happen right now if called incorrectly). But when the sim-objects are loaded on the first tick, this should not happen anymore.

CR1337: OK, then go for it :)

CR1337 (approved):
:+1:

PR #288, opened 2023-04-13
Feature/261 random schedule strategy

### LINKED ISSUE #261: Random Schedule Strategy
Implement Schedule Strategy that allows to spawn randomly.

## Acceptance Criteria
- [ ] Implement `RandomScheduleStrategy`
- [ ] Update `ScheduleConfiguration`
- [ ] Attributes
    - [ ] `start_tick`
    - [ ] `end_tick`
    - [ ] `probability_per_1000`

CR1337:
Fixes #261 

I implemented the `RandomScheduleStrategy`. It is able to spawn trains based on a probability value. It also takes an optional seed for random number generation.

Also I put `start_tick` and `end_tick` in the abstract `ScheduleStrategy` because every strategy should have it.

See [Wiki](https://github.com/BP2022-AP1/bp2022-ap1/wiki/Spawner)

## PR checklist


src/schedule/schedule_configuration.py
```diff
@@ -26,13 +31,18 @@ def _make(self, data: dict) -> "ScheduleConfiguration":
             """
             return ScheduleConfiguration(**data)
 
-    schedule_type = TextField(null=False)
-    strategy_type = TextField(null=False)
+    schedule_type = TextField()
+    strategy_type = TextField()
+
+    strategy_start_tick = IntegerField(null=True)
+    strategy_end_tick = IntegerField(null=True)
```
Lietze, line 38: You said in the pr comment, that every strategy should have a start and an end tick. If so, I don't understand the (null=True). Or did you mean that every type of strategy should have those fields?

src/schedule/schedule_strategy.py
```diff
@@ -20,10 +31,11 @@ def from_schedule_configuration(cls, schedule_configuration: ScheduleConfigurati
         """
         raise NotImplementedError()
 
-    @abstractmethod
     def should_spawn(self, tick: int) -> bool:
```
Lietze, line 34: Maybe this method should be renamed to 'is_between_start_and_end_tick' and 'should_spawn' should stay abstract. Do you think, that this will ever be the only check for any strategy to determine if it should spawn?

src/schedule/regular_schedule_strategy.py
```diff
@@ -14,20 +14,21 @@ def from_schedule_configuration(cls, schedule_configuration: ScheduleConfigurati
         """
```
Lietze, line 16: I also noticed, that an additional typehint could be added just like in random-strategy

tests/schedule/test_schedule_configuration.py
```diff
@@ -40,21 +40,21 @@ def test_deserialization(self, obj: dict):
                 "schedule_type": "TrainSchedule",
                 "strategy_type": "RegularScheduleStrategy",
                 "train_schedule_train_type": "cargo",
-                "regular_strategy_start_tick": 10,
+                "strategy_start_tick": 10,
```
Lietze, line 42: Why do you not define end_tick here? And you don't use the other new fielsd for the random strategy.

tests/schedule/test_train_schedule.py
```diff
@@ -41,7 +41,7 @@ def schedule_id(self) -> str:
     @pytest.fixture
     def schedule(self, platform_ids: list[str], schedule_id: str):
         strategy = RegularScheduleStrategy(
-            start_tick=self.START_TICK, frequency=self.FREQUENCY
+            start_tick=self.START_TICK, end_tick=None, frequency=self.FREQUENCY
```
Lietze, line 44: I would prefere if you used end_tick, so that the funtionality of that also gets tested.

tests/schedule/test_regular_schedule_strategy.py
```diff
@@ -8,7 +8,7 @@ class TestRegularScheduleStrategy:
 
     @pytest.fixture
     def regular_strategy(self) -> RegularScheduleStrategy:
-        return RegularScheduleStrategy(start_tick=1000, frequency=100)
+        return RegularScheduleStrategy(start_tick=1000, end_tick=None, frequency=100)
```
Lietze: Again, I would prefere you using end_tick to also test if that is working.

tests/spawner/test_spawner.py
```diff
@@ -36,14 +36,14 @@ def setup_method(self):
             schedule_type="TrainSchedule",
             strategy_type="RegularScheduleStrategy",
             train_schedule_train_type=self.TRAIN_TYPES[0],
-            regular_strategy_start_tick=self.START_TICKS[0],
+            strategy_start_tick=self.START_TICKS[0],
```
Lietze, line 39: Also here you may want to also use end_tick. Or maybe I don't get why that is not what you want, that is also totaly possible.

Lietze (commented):
This looks great. I think the new stuff is great, but I found a few things in the old strategy that should be changed.

src/schedule/schedule_configuration.py
CR1337: Oh sorry. My statement was a bit misleading. I meant every type of strategy should be able to receive as `start_tick` and a `end_tick` during its construction. But they can be left empty. Then `start_tick` will default to `0` and `end_tick` will default to "infinity".

I hope that makes it clearer :)

src/schedule/schedule_strategy.py
CR1337: The inheriting classes also implement `should_spawn` and then call `super().should_spawn`. So every class adds its own logic and implements it at its own inheritance level.

CR1337: If I have to add another check I simply add it the `should_spawn` of the responsible class.

src/schedule/regular_schedule_strategy.py
CR1337: Done :)

tests/schedule/test_schedule_configuration.py
CR1337: I added an `end_tick` and also some objects for the random strategy.

tests/schedule/test_regular_schedule_strategy.py
CR1337: I added a test.

tests/schedule/test_train_schedule.py
CR1337: As `end_tick` is tested in both strategies it is not necessary here in my opinion. Contrary setting it to None tests that it also works when `end_tick` is None.

tests/spawner/test_spawner.py
CR1337: See above.

As `end_tick` is tested in both strategies it is not necessary here in my opinion. Contrary setting it to None tests that it also works when `end_tick` is None.

tests/schedule/test_regular_schedule_strategy.py
```diff
@@ -22,6 +22,18 @@ def test_not_spawning_before_start_tick(
...
+        "strategy,tick", (("regular_strategy", tick) for tick in [20000])
+    )
+    def test_not_spawning_after_end_tick(
+        self,
+        strategy: RegularScheduleStrategy,
+        tick: int,
+        request: pytest.FixtureRequest,
+    ):
+        strategy = request.getfixturevalue(strategy)
+        assert not strategy.should_spawn(tick=tick)
```
Lietze:
```suggestion
    @pytest.mark.parametrize(
        "tick", [20000]
    )
    def test_not_spawning_after_end_tick(
        self,
        regular_strategy: RegularScheduleStrategy,
        tick: int
    ):
        assert not regular_strategy.should_spawn(tick=tick)
```
You could change this test (and those like it) to use the fixture directly. I think this would make it easier to understand, because the parametrizing has only one parameter and you don't need to use FixtureRequest anymore.

Lietze (approved):
Thank you for answering my questions. I took another look at the test in ```test_regular_schedule_strategy``` and think they could be made easier. This may not be the issue for that though, so I will approve anyway. Other than that this looks good.

tests/schedule/test_regular_schedule_strategy.py
CR1337: Good idea. Thanks :)

PR #313, opened 2023-04-17
added tests that are based on fault_config tests

### LINKED ISSUE #271: Update Interlocking Configuration
## User stories
As a developer, I want the interlocking configuration to be implemented the same way as other database classes in order to understand how it works and how it is tested.

## Wiki Reference
- [the database class diagram](https://github.com/BP2022-AP1/bp2022-ap1/wiki/Architecture#database-class-diagram)

## Acceptance Criteria
- [x] add it to constants.py
- [x] update tests to be equal to the others

Lietze:
Fixes #271 

I copied and modified the test from the test_table_fault_configurations.py

## PR checklist


tests/tests_interlocking/test_interlocking_configuration.py
```diff
@@ -1,8 +1,55 @@
...
+        assert (
+            InterlockingConfiguration.select()
+            .where(InterlockingConfiguration.id == obj.id)
+            .first()
+            == obj
+        )
+
+    def test_serialization(self, object_as_dict: dict):
+        """Test that an object of a class can be serialized."""
+        obj = InterlockingConfiguration.create(
```
CR1337: You are not serializing here.

You create an object using the peewee constructor. Then you check whether the resulting `InterlockingConfiguration` object has the same content as the dict. 

What you want to test instead is whether the conversion of the object to a dict works. Therefore you could call `obj.to_dict`. Then you could test if the resulting dict has the same content as the original dict.

tests/tests_interlocking/test_interlocking_configuration.py
```diff
@@ -1,8 +1,55 @@
...
+            **object_as_dict,
+        )
+
+        for key in object_as_dict.keys():
+            assert getattr(obj, key) == object_as_dict[key]
+
+    def test_deserialization_full_dict(self, object_as_dict: dict):
+        """Test that an object of a class can be deserialized."""
+        obj = InterlockingConfiguration.Schema().load(
```
CR1337: This works completely fine. But is there a reason not to use `InterlockingConfiguration.from_dict` ? 🤔

CR1337 (changes requested):
Looks good. There are just some things in your tests :)

tests/tests_interlocking/test_interlocking_configuration.py
Lietze: I have changed it. The only reason was copy and paste. :wink:

tests/tests_interlocking/test_interlocking_configuration.py
Lietze: Thank you for the feedback, I still don't understand peewee+marshmallow 100%, but I have changed it now. @instant-sky, this comment and the one below also applies to your tests. I think you should change this too.

CR1337 (approved):
Looks good. Thanks for the changes :)

PR #340, opened 2023-04-18
Add insert_priority_changed

Lietze:
Fixes #339 

## PR checklist

- [ ] Acceptance criteria fulfilled
- [ ] Additional features are tested
- [ ] Docs (code, wiki & diagrams) updated
- [ ] Breaking changes anounced
- [ ] Added classes that inherit from `BaseModel` to `src.constants.tables`
- [ ] Dev-branch has been merged into local branch to resolve conflicts
- [ ] Tests and linter have passed AFTER local merge
- [ ] Another dev reviewed and approved

PR #341, opened 2023-04-18
Add functionality to TrainPrioFault

### LINKED ISSUE #317: Add functionality to `TrainPrioFault`
## User stories
As a developer I want functionality in `TrainPrioFault` in order to manipulate the priority of a train.

## Acceptance Criteria
- [x] implement `inject_fault`
- [x] implement `resolve_fault`
- [x] write tests

instant-sky:
Fixes #317 

## PR checklist

- [ ] ~Breaking changes anounced~
- [ ] ~Added classes that inherit from `BaseModel` to `src.constants.tables`~


instant-sky:
The tests are currently failing because this PR is dependent on #340

src/fault_injector/fault_types/train_prio_fault.py
```diff
@@ -1,27 +1,67 @@
...
         """inject TrainPrioFault into the given component
 
         :param tick: the simulation tick in which inject_fault was called
         :type tick: Integer
         """
-        # - get train object
-        # - save the current prio of the train in old_prio
-        # - set train prio to new_prio
-        raise NotImplementedError()
+        self.train: Train = [
```
antonykamp, line 41: Are you entirely sure that you will find the train? Maybe you want to check for it first.

src/fault_injector/fault_types/train_prio_fault.py
```diff
@@ -1,27 +1,67 @@
...
+            train
+            for train in self.wrapper.trains
+            if train.identifier == self.configuration.affected_element_id
+        ][0]
+        # pylint: enable=duplicate-code
+        self.old_prio = self.train.train_type.priority
+        self.train.train_type.priority = self.configuration.new_prio
```
antonykamp, line 48: Are you sure that you change the priority of a single train only and not every train with the same type? 🙃

antonykamp (commented):
Just some questions, noting to worry about 🙈

src/fault_injector/fault_types/train_prio_fault.py
```diff
@@ -1,27 +1,67 @@
...
         """
-        # - get train object
-        # - save the current prio of the train in old_prio
-        # - set train prio to new_prio
-        raise NotImplementedError()
+        self.train: Train = [
+            train
+            for train in self.wrapper.trains
+            if train.identifier == self.configuration.affected_element_id
+        ][0]
```
Lietze, line 45: Why is [0] needed? Can there ever be two trains with the same id?

src/fault_injector/fault_types/train_prio_fault.py
instant-sky: Ah that is a good point 👇. Since I've missed that in the other Faults as well I would fix that with an extra issue.

antonykamp: Perfect 👍 Please reference the new issue here ;)

src/fault_injector/fault_types/train_prio_fault.py
```diff
@@ -1,27 +1,67 @@
...
+        # pylint: enable=duplicate-code
+        self.old_prio = self.train.train_type.priority
+        self.train.train_type.priority = self.configuration.new_prio
+
+        self.interlocking.insert_train_priority_changed(self.train.identifier)
```
Lietze, line 50:
```suggestion
        self.interlocking.insert_train_priority_changed(self.train)
```
```insert_train_priority_changed``` takes trains as arguments, so this would be better.

File: src/fault_injector/fault_types/train_prio_fault.py (hunk 2 above)
Comment (instant-sky) [reply src/fault_injector/fault_types/train_prio_fault.py:48]: At first I as well thought, that this would change the `max_speed` of the entire type and thus on every existing train of this type. But apparently the behaviour of `train_type` is different from what you would expect 🤷‍♂️. This at least is what @SaturnHafen told me.

File: src/fault_injector/fault_types/train_prio_fault.py [hunk 5]
```diff
@@ -1,27 +1,67 @@
...
         """resolves the previously injected TrainPrioFault
 
         :param tick: the simulation tick in which resolve_fault was called
         :type tick: Integer
         """
-        # - get train object
-        # - set the train prio to old_prio
-
-        raise NotImplementedError()
+        self.train.train_type.priority = self.old_prio
```
Lietze, line 65: I think it is needed to check for ```None``` here, as the train could have left the simulation.

src/fault_injector/fault_types/train_prio_fault.py
Lietze, line 41: I think you need to check for ```None``` as the train may not yet have been spawned.

src/fault_injector/fault_types/train_prio_fault.py
instant-sky: I looked it up again, train_type just stores the Metadata of that train 
```
class Train(SimulationObject):
    """A train driving around in the simulation."""

    # pylint: disable=too-many-instance-attributes

    class TrainType(SimulationObject):
        """Metadata about a specific train"""

        _max_speed: float = None
        _priority: int = None
        _name: str = None

```

tests/fault_injector/test_train_prio_fault.py
```diff
@@ -0,0 +1,79 @@
...
+        # pylint: disable-next=unused-argument
+        combine_train_and_wrapper,
+    ):
+        train.train_type.priority = 5
+        with pytest.raises(NotImplementedError):
+            train_prio_fault.inject_fault(tick)
+        assert train.train_type.priority == train_prio_fault.configuration.new_prio
+        with pytest.raises(NotImplementedError):
+            train_prio_fault.resolve_fault(tick)
+        assert train.train_type.priority == train_prio_fault.old_prio == 5
```
Lietze, line 79: I think you should add a test for changing Prio before the train is spawned and one for changing Prio after the train is spawned.

Lietze (commented):
Good job! You should think about the behavior if the train is not spawned yet or has despawned already. This then also needs to be applied to the ```train_speed_fault```. Other than that it looks good.

src/fault_injector/fault_types/train_prio_fault.py
antonykamp: Top 👍 Thank you!

src/fault_injector/fault_types/train_prio_fault.py
instant-sky: I hope not, but I get a list from the for loop above

src/fault_injector/fault_types/train_prio_fault.py
instant-sky: Yes, this is an issue in all of my faults, that gets fixed with #346

src/fault_injector/fault_types/train_prio_fault.py
instant-sky: Hm, even if the train left the simulation, shouldn't I still have the corresponding train object? But you are right, some `None` checks are missing, this issue is tackled in #353

src/fault_injector/fault_types/train_prio_fault.py
instant-sky: Yes, issue #353 fixes that. Since this problem exists in all of my faults I wanted to change that in all of them at once

tests/fault_injector/test_train_prio_fault.py
instant-sky: Good point, since they would fail right now and the `None` checks are in a different issue I think adding the test there is better

antonykamp (approved):
looks good to me for now :)

src/fault_injector/fault_types/train_prio_fault.py
Lietze: ok that is good

PR #360, opened 2023-04-20
Documentation: Add short description about user workflow to openAPI3

### LINKED ISSUE #358: Add user workflow with REST API to wiki
## User stories
As a future user, I want to know the main principles of the API to use it correctly.

## Acceptance Criteria
- [x] Add description of user workflow to openAPI3 definition

antonykamp:
Fixes #358 

Add a short description of user workflow to openAPI3 definition file.

## PR checklist


postman/schemas/index.yaml
```diff
@@ -1,8 +1,13 @@
 openapi: 3.0.0
 info:
   description:
-    An API for simulation and analysis of train traffic on the LEAG rail
-    network
+    A REST API for simulations and analysis of train traffic on the LEAG rail
+    network. Create one of many component configurations, 
+    for example, for defining the interlocking, train schedules, and faults.
+    You can't edit an already specified component configuration but can delete it, including its connections to simulation configurations.
```
CR1337: What happens when you delete it? 
Does the simulation configuration then have a NULL reference or is the deletion cascaded?
Both doesn't sound very desirable 🤔

antonykamp: Good question! 
Deleting a component configuration while a connection to a simulation configuration still exists shouldn't be possible, IMO. You have to delete the connections first ;)

CR1337: But what happens when you delete the connections?

What if a run already happened using a simulation configuration. Then we delete a connection to the spawner configuration. Now we want to do some data science on the run. But the information about the spawner is gone. This shouldn't happen. Isn't it better to also forbid deleting?

CR1337: Or we cascade the deletion. That means to also delete all simulation and run using the deleted item.

antonykamp: > Then we delete a connection to the spawner configuration.

That's not allowed 😇 It'll fail.

CR1337: Ok, and what deletion is allowed?
I believe I didn't get it yet 😄

antonykamp: I propose: 
- You can delete a component configuration if no connection to a simulation configuration exists.
- You can't delete a connection to a component configuration if at least one run exists for the simulation configuration.
- You can't delete a simulation configuration if a connected run exists. You have to delete the run, including the logs, first. After that, you can delete a simulation configuration, including the connections to the component configuration but not the configuration itself.

CR1337: That sounds good. My inner database-extremist is satisfied :D

CR1337 (approved):
:+1:

PR #370, opened 2023-04-21
Docs: Generate REST API documentation with sphinx

### LINKED ISSUE #335: Generate and host API documentation with sphinx
## User story
As a developer and user I want to have a public API documentation online.

## Acceptance criterium
- [x] Install sphinx
- [x] Add https://github.com/sphinx-contrib/openapi
- [ ] Host documentation

antonykamp:
Fixes #370

This PR install sphinx first to generate a documentation from the openAPI3 definition file.
It adds a tool to generate the documentation and a theme.
Additionally, I added some poe commands and host the documentation on readthedocs.

https://bp2022-ap1.readthedocs.io/en/doc-335-generate-and-host-api-documentation-with-sphinx/

## PR checklist


docs/requirements.txt
SaturnHafen, line 1: Why do we need seperate dependencies? Why can't we use the `pyproject.toml` dependencies?

docs/Makefile
SaturnHafen, line 1: Why do we need the `Makefile` and the `make.bat` if we have the `poe` commands?

SaturnHafen (commented):
I like that we get compiled docs! Here are some things that bother me: I don't like that we are using a separate build system (`make`) and dependency-specification (`requirements.txt`) for the docs. Doing a quick search brought something like [this](https://tomasfarias.dev/posts/sphinx-docs-with-poetry-and-github-pages/). Could we utilize that somehow? Also you defined `poe` commends that seem to invoke the sqhinx-builder directly, why do we need the makefiles then? Additionally: Could you please add a github-action that deploys the docs to github-pages, when we commit on main/dev (or add a ticket for that)...

docs/Makefile
antonykamp: I'm not happy this either. Sphinx generated this file and needs it to generate the htmls.

> Also you defined poe commends that seem to invoke the sqhinx-builder directly, why do we need the makefiles then

The sphinx builder uses the make file 🙃

docs/requirements.txt
antonykamp: readthedocs needs the `requirements.txt` file. We could switch to GitHub pages. I think it's a good idea because otherwise, I would be the only person to manage the readthedocs project. 👍

docs/Makefile
SaturnHafen: Now I'm confused. If I read the [docs](https://www.sphinx-doc.org/en/master/man/sphinx-build.html) correctly, the only command that we should need to execute is (we are using the not-make html builder, as you did in the poe commands)
```sh
sphinx-build -b html  "docs/source" "docs/build/html"
```

antonykamp: I'll try it again without the make file, but I found this too:
https://www.sphinx-doc.org/en/master/man/sphinx-build.html#makefile-options

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: poetry install --with docs
+      - name: Build documentation
+        run: |
+          poetry run poe docs-build
+      - name: Setup Pages
+        uses: actions/configure-pages@v3
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v1
+        with:
+          # Upload entire repository
```
SaturnHafen:
```suggestion
          # Upload docs output
```

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,54 @@
+# Simple workflow for deploying static content to GitHub Pages
+name: Deploy static content to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["dev"]
```
SaturnHafen, line 7: We only have a unstable-deployment, no "stable" documentation, right?

SaturnHafen (approved):
Looks good to me, we might want to think about a stable release documentation vs. a unstable development documentation

.github/workflows/deploy-docs.yml
antonykamp: It's an interesting discussion. I think as long as we don't have a real deployment, we should keep this one 🙃

PR #403, opened 2023-04-27
Feature: Add user token validation decorator

### LINKED ISSUE #162: Add Token validation decorator
## User stories
As an admin of the REST API I want that it's secure and only approved user can use the API in order to prevent downtimes.

## Dependencies
Depends on: 
 - #99 
 - #100

## (Wiki) References
- https://flask.palletsprojects.com/en/2.2.x/patterns/viewdecorators/

## Acceptance Criteria
- [ ] Add decorator
- [ ] Add fetching user information from header
- [ ] Add tests

antonykamp:
Fixes #162 

Add decorator for flask api functions to extract, validate and forward user token.

## PR checklist


src/api/token.py
```diff
@@ -2,16 +2,18 @@
...
+from src.api.decorators import token_required
 from src.schemas import model
 
 bp = Blueprint("token", __name__)
 
 
 @bp.route("/token", methods=["post"])
-def create_token():
+@token_required
+def create_token(token):
```
CR1337, line 13: Can you explain why a token is required to create a token? Or in other words, how would you create a token when you don't have one initially? 🤔

src/implementor/token.py
```diff
@@ -4,10 +4,11 @@
 import json
 
 
-def create_token(body):
+def create_token(body, token):
```
CR1337, line 7: Same question here. Why does token creation need a token?

CR1337: Also as this is not implemented yet, how do you plan to generate a token. On this depends whether your decorator is safe or not.

CR1337 (commented):
I have some questions, but maybe I just don' t fully understand your code.

src/api/token.py
antonykamp: Of course! Good question. Only admins should be able to create tokens. Therefore the first admin token has to be inserted into the database by hand 🙃

src/implementor/token.py
antonykamp: See above 😇

src/api/token.py
CR1337: Ok, that makes sense :)

src/implementor/token.py
CR1337: What itches my brain is that you already have code to check a token but there is not code to generate one yet. Or did I overlook something?

I just wanted to get a feeling for the safety of our tokens. But I can also do that later when you plan to implement generation in another PR 🙂

CR1337 (approved):
Then go for it :+1:

PR #412, opened 2023-05-03
Refactor: Usage of marshmellow schema

### LINKED ISSUE #411: Refactor usage of marshmellow schema
## User stories
As a developer of the REST API, I want to have an implementation of marshmallow schemas where I can instantiate a configuration instance from a JSON string including references. That'll help me to have a uniformly implementation of the endpoints.

## (Wiki) References
- https://github.com/BP2022-AP1/bp2022-ap1/wiki/ORM-&-(de)serialization

## Acceptance Criteria
- [x] Split BaseModel into BaseModel and SerializableBaseModel that interhits from BaseModel
- [x] Use SerializableBaseModel if needed
- [x] Remove Schema if not needed anymore
- [x] Update tests
- [x] Update wiki

antonykamp:
Fixes #411 

Only use marshmellow schemas if needed directly from the rest API.
Therefore I had to create a second serializable BaseModel and use it or remove the `Schema`-classes.

## PR checklist


antonykamp:
I'll add the discussed new implementation of `_make` when implementing the POST endpoints @CR1337 😇

src/base_model.py
```diff
@@ -70,3 +58,20 @@ def to_dict(self) -> dict:
...
         """
         return self.Schema().dump(self)
+
+    class Schema(marsh.Schema):
+        """The marshmallow schema all model schemas have to inherit from."""
+
+        id = marsh.fields.UUID(dump_only=True)
+        created_at = marsh.fields.DateTime(format="iso", dump_only=True)
+        updated_at = marsh.fields.DateTime(format="iso", dump_only=True)
+
```
CR1337, line 68:
```suggestion
@abc.abstractmethod
def _dump(self, obj: 'SerializableBaseModel') -> dict:
    raise NotImplementedError()

@marsh.pre_dump         
def dump(self, obj: 'SerializableBaseModel', **_) -> dict:
    return self._dump(obj)
```

As you mentioned via Slack you want the funktionality of `make` the other way around. You can try my suggestion. Please validate if it really works.

You can read more about it [here](https://marshmallow.readthedocs.io/en/stable/extending.html) and [here](https://marshmallow.readthedocs.io/en/stable/marshmallow.decorators.html#marshmallow.decorators.pre_dump).

CR1337: You would then override `_dump` in the subclasses to specify the serialization behavior.

CR1337 (commented):
All my tests 😭

Looks good so far. I left a suggestion for customizable serialization behavior :)

src/base_model.py
antonykamp: Thank you very much for the research and sorry regarding your tests 💀 
I'd stay with overwriting `to_dict()` function if needed. We'll have some classes where `Schema().dump()` is completely appropriate and sometimes not. That's why I won't add an abstract method `dump()` with a decorator and the abstract method `_dump()`. 🙃

antonykamp: If I understand it correctly, `_dump()` would take the object as input and return the dictionary?

CR1337: It's ok I didn't like the tests anyway :D

My inner linter doesn't like that serialization is achieved with a different method then deserialization. But considering our time constraints that is totally ok :)

CR1337 (approved):
:+1:

src/base_model.py
antonykamp: > My inner linter doesn't like that serialization is achieved with a different method then deserialization.

same here 🤔

CR1337: I'm ok with ignoring it

antonykamp: I'm unsure what happens when the dumping method doesn't find `platform_ids` list in the class serialize.

CR1337: There are platform ids. A `ScheduleConfigruation`-object has a  `train_schedule_platform_references` field. It gets that field by the Foreign Key definition in `ScheduleConfigurationXSimulationPlatform`.

antonykamp: Yes I understand, but it's not called `platform_ids` :/ I'm worried that the Schema is confused, if it doesn't find anything at `platform_ids`. arggghhh
And  train_schedule_platform_references doesn't hold the ids. Instead it holds the object containing the id ;)

CR1337: Yes you are right. I'm sure there is a way to make it elegant and "right". But that would definitely take more time. In the end it is your decision. I am fine with the faster alternative.

antonykamp: Alright. I'll try something later, but for now i'll leave it this way 🙃 
Thank you for the discussion, your research and comments! 💚

PR #422, opened 2023-05-05
Feature: Add functionality to /run route

### LINKED ISSUE #196: Add functionality of `/run` route
## User stories
As a user I want start a new run and get all my runs in order to administrate them.

## Dependencies
Depends on: 
 - #99 
 - #102 
 - #100
 - #141 

## Acceptance Criteria
- [x] Implement POST `/run` path
- [x] Implement GET `/run` path
- [x] Add to collections
- [ ] Add newman tests

antonykamp:
Fixes #196 

This PR introduces the functionality of the GET and POST /run route and starts a run.
Therefore I had to remove the celery wrapper of the flask app and introduce a single celery instance in the communicator. 
That means I had to adjust the Thread-based implementation of the communicator with a celery-based one.
Additionally, I've added the fields "process_id" to reference the process of a run.

## PR checklist


antonykamp:
The not-covered lines in coveralls are covered 🤔

compose/celery/Dockerfile
```diff
@@ -1,9 +1,15 @@
 FROM python
 WORKDIR /app
 
+# RUN apt-get update
+# RUN apt-get install -y software-properties-common
+
+#RUN add-apt-repository ppa:sumo/stable
+RUN apt-get update
+RUN apt-get install -y sumo
```
SaturnHafen: I am not sure why every container needs sumo... Could you explain that to me, please?
Additionally: please don't comment out the ppa, because then we will get a version mismatch between traci and sumo (the default debian package index is outdated)
(Applies to the other images as well)

src/communicator/communicator.py
```diff
@@ -50,34 +53,78 @@ def __init__(
...
         """Creates a new Communicator object"""
-        Thread.__init__(self)
         self._configuration = sumo_configuration
         self._port = sumo_port
         self._components = components if components is not None else []
         self._max_tick = max_tick
 
-    def run(self):
+    def run(self) -> str:
+        process = self._run.delay(
```
SaturnHafen: Add a method comment please :)

src/component.py
SaturnHafen, line 1: This should be placed next to the tests, shouldn't it?

make_celery.py
```diff
@@ -1,4 +1,4 @@
-from src import create_app
+from src.communicator.communicator import celery
 
-flask_app = create_app()
-celery_app = flask_app.extensions["celery"]
+if __name__ == "__main__":
+    celery.start()
```
SaturnHafen, line 4: Why is the communicator responsible for creating the celery context?

docker-compose.yaml
```diff
@@ -48,18 +48,21 @@ services:
...
+      - 6379:6379
 
   celery-worker:
     build:
       context: .
       dockerfile: ./compose/celery/Dockerfile
     env_file:
       - .env.shared
+    environment:
+      - C_FORCE_ROOT=true
```
SaturnHafen, line 61: This sounds sketchy, why do we need that?

docker-compose.yaml
```diff
@@ -48,18 +48,21 @@ services:
...
     volumes:
       - .:/app
     depends_on:
       - redis
-      - postgresql
```
SaturnHafen, line 62: If we don't need the database anymore, please remove it also from the top :)

.env.test
```diff
@@ -5,6 +5,11 @@ DATABASE_PASSWORD=root
...
+
 SECRET_KEY=test
+
 GF_SECURITY_ADMIN_PASSWORD=admin
-GRAFANA_HOST=grafana-dev
\ No newline at end of file
+GRAFANA_HOST=grafana-dev
+
+CELERY_BROKER_URL=redis://localhost:6379/0
+CELERY_RESULT_BACKEND=redis://localhost:6379/0
```
SaturnHafen, line 15: Does localhost work (or should you use the docker-dns-names). Docker+loacalhost typically don't work that well together

src/communicator/communicator.py
```diff
@@ -1,36 +1,39 @@
...
 from src.component import Component
 
-
-class Communicator(Thread):
+celery = Celery(
+    "proj",
+    broker=os.environ["CELERY_BROKER_URL"],
+    backend=os.environ["CELERY_RESULT_BACKEND"],
+)
+celery.conf.event_serializer = "pickle"  # this event_serializer is optional. somehow i missed this when writing this solution and it still worked without.
```
SaturnHafen: Why this comment?

src/communicator/communicator.py
```diff
@@ -50,34 +53,78 @@ def __init__(
...
+        self: Task,
+        max_tick: int,
+        components_pickle: bytes,
+        configuration: str,
+        port: int,
+    ):
         """Starts sumo (no gui) and connects using traci. The connection has the `default` label."""
 
-        traci.start([checkBinary("sumo"), "-c", self._configuration], port=self._port)
+        components = pickle.loads(components_pickle)
```
SaturnHafen, line 96: I am confused: Why do we need to pass the components serialized?

src/communicator/communicator.py
```diff
@@ -50,34 +53,78 @@ def __init__(
...
+        update_state()
 
-    def stop(self):
+    @classmethod
+    def stop(cls, process_id: str):
         """Stopps the simulation requesting a simulation step.
         At most, one simulation step will happen after this request"""
-        self._stopped = True
+        process = AsyncResult(process_id)
+        process.revoke()
```
SaturnHafen, line 134: You may need to change the comment regarding the delay of the cancellation

src/implementor/run.py
```diff
@@ -3,41 +3,196 @@
...
+        reference
+    ) in simulation_configuration.platform_blocked_fault_configuration_references:
+        platform_blocked_fault_config = reference.platform_blocked_fault_configuration
+        fault = PlatformBlockedFault(
+            platform_blocked_fault_config,
+            logger,
+            object_updater,
+            interlocking_disruptor,
+        )
+        communicator.add_component(fault)
```
SaturnHafen, line 123: One could extract this to a separate method, but I think that is not necessary

tests/communicator/test_communicator.py
```diff
@@ -29,46 +30,40 @@ def close(*args, **kwargs):  # pylint: disable=unused-argument
...
-):  # pylint: disable=unused-argument
+def test_simulation_runs(mock_traci):  # pylint: disable=unused-argument
     communicator = Communicator()
-    communicator.start()
-    sleep(0.1)
-    communicator.stop()
-    assert communicator.progress > 0
+    run_id = communicator.run()
+
+    while Communicator.state(run_id) != "PROGRESS":
```
SaturnHafen, line 37: Where does this `"PROGRESS"` come from? Is that a celery thing?

tests/implementor/test_run.py
```diff
@@ -0,0 +1,103 @@
...
+    def test_get_all_simulation_ids_option(
+        self,
+        token,
+        empty_simulation_configuration,
+        another_empty_simulation_configuration,
+    ):
+        run = Run.create(simulation_configuration=empty_simulation_configuration)
+        another_run = Run.create(
+            simulation_configuration=another_empty_simulation_configuration
+        )
```
SaturnHafen, line 28: Should these runs be created in a fixture?

SaturnHafen (commented):
I have some questions and comments on your changes, but it looks really good. With this implemented, I am sure that we will get the MVP to work in time :).

compose/celery/Dockerfile
antonykamp: > I am not sure why every container needs sumo... Could you explain that to me, please?

Of course! Celery runs in its own service/conainer where the started processes run too. That means, the content of the method `_run` runs in this service, including [this line](https://github.com/BP2022-AP1/bp2022-ap1/pull/422/files#diff-f429acc14efc274f7ba6f9aa41f241c1c5cfa0e55b1b54e0226359610b12c60cR81), where we check for sumo-binaries. Therefore we have to install Sumo because the simulation will be run in the started celery process.

> please don't comment out the ppa, because then we will get a version mismatch between Traci and sumo (the default debian package index is outdated)

I tried 😇 Maybe you can help me out here. I couldn't add the repository, because of the following error: `repository/PPA does not have a Release file?`. Do you have another idea than building sumo manually?

src/component.py
antonykamp: That's not possible. When placing it next to the test, `pickle.loads` cannot find the class anymore. The reason is that the test folder is not being bundled into the application, and Celery runs a bundled version of the application. That's why I had to place it there.

make_celery.py
antonykamp: Should I place it into its own file? That's probably a good idea. I put it there because it's part of the communicator's functionality to create new processes. Other files shouldn't have access to it.

docker-compose.yaml
antonykamp: I'm using pickle to serialize the ComponentObject. Celery recommends not doing it, because it's not 100% secure. The alternative is to serialize with JSON. Therefore we have to implement serialization and deserialization for every component. Because that would take some time, I decided to use the unsecured and fast way for the MVP. 
To do that, we need to enable `C_FORCE_ROOT`.

docker-compose.yaml
antonykamp: We still use it in the services "production" and "development" 😇 That's why I want to keep the service. It's just not needed for the celery worker

src/communicator/communicator.py
antonykamp: I copied a solution from another project and forgot to remove the comment 😇 Sorry! I'll remove it

src/communicator/communicator.py
antonykamp: We have to pass a serialized version of the components because of the functionality of celery. Celery runs the started processes in a new service/container, "celery-worker". To run the method call as a new process, celery passes the function arguments via an internet protocol. It only allows primitive datatypes, so we must serialize the components.

src/communicator/communicator.py
antonykamp: Which comment do you mean exactly? 🙃

tests/communicator/test_communicator.py
antonykamp: Yes! Absolutely 😇

.env.test
antonykamp: Yes 👍 If we run tests, the process with the test runs outside of the docker container. That's why we have to expose the port of the Redis database and use the localhost. The development and production container run inside docker and use the docker-dns names. Lucas and I considered running the tests inside docker but decided against it because it's not an infrastructure.

src/communicator/communicator.py
antonykamp: Thanks for the hint, I'll add it directyl!

tests/implementor/test_run.py
antonykamp: I'm not a big fan of fixtures using this way. I have the rule, that I use fixtures as infrastructure I can use. But I don't do assertions with them, because something can change an break in tests, you know?

src/communicator/communicator.py
antonykamp: Found it, thank you 😇

compose/celery/Dockerfile
SaturnHafen: > Therefore we have to install Sumo because the simulation will be run in the started celery process.

Okay, but we have like 5 containers now, why does every one of them need sumo

 > Maybe you can help me out here.

I will look into that

antonykamp: > Okay, but we have like 5 containers now, why does every one of them need sumo

Only the development, production, and celery components install sumo the moment. Because sumo only runs in the celery worker, I can remove it from the development and production component. Thanks for the hint. Can we remove it from the github action script too? 

> I will look into that

Thank you 😇

SaturnHafen: > I will look into that

The problem is that the python container is based on `debian-bullseye`, but sumo does not distribute binaries for that version (see [here](https://launchpad.net/~sumo/+archive/ubuntu/stable) ~> *Technical details about this PPA* ~> *Choose your ubuntu version*)). There are at least two options now: 
1. Use an other base image like this (the image does not work because it installs an outdated version of python):
```dockerfile
FROM debian:bullseye
WORKDIR /app

RUN echo "deb https://ppa.launchpadcontent.net/sumo/stable/ubuntu bullseye main" >> /etc/apt/sources.list.d/sumo.list
RUN apt-get update && apt-get install -y sumo python3 python3-pip

RUN pip install poetry

COPY . /app
RUN poetry install
CMD poetry run celery -A make_celery worker --loglevel=info
```
2. build sumo ourselves (will probably be a pain)

SaturnHafen (approved):
It looks fine now, please fix the dockerfile as sumo will probably not run with this major version mismatch.

PR #430, opened 2023-05-08
log events in the interlocking component

### LINKED ISSUE #418: Log Events in the interlocking
## User stories
As a logger, I want certain events to be logged in order to provide information about those events.

## (Wiki) References
- [List of availabe logging functions](https://github.com/BP2022-AP1/bp2022-ap1/wiki/Logging#log-creator)

## Acceptance Criteria
- [x] log whenever a fahrstrasse is created
- [x] log whenever a fahrstrasse is removed
- [x] log when a signal is set by the infrastructure provider
- [x] log when a train enters a block section
- [x] log when a train leaves a block section

Lietze:
Fixes #418

## PR checklist

### DIFF
--- a/src/interlocking_component/infrastructure_provider.py
+++ b/src/interlocking_component/infrastructure_provider.py
@@ -21,7 +21,7 @@ class SumoInfrastructureProvider(InfrastructureProvider):
     ):
         super().__init__()
         self.route_controller = route_controller
-        self.route_controller.simulation_object_updating_component.infrastructur_provider = (
+        self.route_controller.simulation_object_updating_component.infrastructure_provider = (
             self
         )
         self.logger = logger
--- a/src/interlocking_component/infrastructure_provider.py
+++ b/src/interlocking_component/infrastructure_provider.py
@@ -81,7 +81,7 @@ class SumoInfrastructureProvider(InfrastructureProvider):
 
         self.route_controller.maybe_set_fahrstrasse(train, edge)
 
-    def train_drove_off_track(self, edge: Edge):
+    def train_drove_off_track(self, train: Train, edge: Edge):
         """This method calls tds_count_out with the track_segment_id of the given edge.
 
         :param edge: The edge the train drove off of
@@ -91,4 +91,4 @@ class SumoInfrastructureProvider(InfrastructureProvider):
         # The interlocking does not have two edges per track, so the -re must be removed if there
         self.tds_count_out(track_segment_id)
 
-        self.route_controller.maybe_free_fahrstrasse(edge)
+        self.route_controller.maybe_free_fahrstrasse(train, edge)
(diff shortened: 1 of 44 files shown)


src/interlocking_component/infrastructure_provider.py
```diff
@@ -64,8 +80,9 @@ def train_drove_onto_track(self, train: Train, edge: Edge):
         self.tds_count_in(track_segment_id)
 
         self.route_controller.maybe_set_fahrstrasse(train, edge)
+        print(train.identifier)
```
LucasDerReisende: maybe remove this print if not needed

LucasDerReisende (approved):
looks good :)

LucasDerReisende:
I am just wondering, why the coverage decreased here by that much?

src/interlocking_component/infrastructure_provider.py
CR1337: Maybe you want to remove this?

src/interlocking_component/route_controller.py
```diff
@@ -194,7 +196,14 @@ def set_fahrstrasse(self, train: Train, edge: Edge):
         new_route = self.router.get_route(edge, train.timetable[0].edge)
         # new_route contains a list of signals from starting signal to end signal of the new route.
 
-        for end_node_candidat in new_route[1:]:
+        route_length = 0
+
+        for i in range(new_route - 1):
```
CR1337: `new_route` is a list of nodes, right?
Then this doesn't work. `range` uses an integer and not an iterable.
Either use `for node in new_route` or if you really need an index use `for i in range(len(new_route))`. 
If you need both use `for i, node in enumerate(new_route)`.

CR1337 (changes requested):
Overall it looks good. Please just have a quick look at my comments :)

src/interlocking_component/infrastructure_provider.py
Lietze: Yes, that slipped through.

Lietze: Yes I did

src/interlocking_component/route_controller.py
Lietze: I changed this

CR1337 (approved):
:+1:

Lietze:
> I am just wondering, why the coverage decreased here by that much?

The only thing I can think about is, that I am now using a MockLogger instead of the real one. Maybe I was testing some code of the logger before? If so, this would not be the place to test that code

LucasDerReisende:
It seems like it's not the Logger, but in `route_controller.py` and `simulation_objects.py`. Could you please have a look at those, because when looking at the coverage difference on dev, it seems like it decreased by 0.4%. If you know the reason and it's intended, I don't think it's a problem, just want to make sure, that we did not remove tests for certain functions.

PR #461, opened 2023-05-12
Event bus

### LINKED ISSUE #455: Implement EventBus
## User stories
As a developer I want to be able to subscribe to events happening in the simuation in order to react to them

## Depends on
- #443 

## (Wiki) References
- [EventBus](https://github.com/BP2022-AP1/bp2022-ap1/wiki/EventBus)

## Acceptance Criteria
- [ ] Implement `EventConsumer` that can subscribe und unsubscribe to events with callbacks
- [ ] Implement `EventBus` that replaces the logger in all components and takes all event notifications
- [ ] Make the `Logger` an `EventConsumer`
- [ ] Implement `Event` and  `EventType` to use as argument for event callbacks

instant-sky:
Fixes #455 

Adds an event bus that enables components to react to events.
For documentation see https://github.com/BP2022-AP1/bp2022-ap1/wiki/EventBus#updated-version-of-more-indirection

## PR checklist

- [ ] ~~Added classes that inherit from `BaseModel` to `src.constants.tables`~~


src/event_bus/event_bus.py
```diff
@@ -0,0 +1,202 @@
...
+        """Cancels the subscription associated to a handle
+
+        :param handle: the handle
+        """
+        self.callbacks.pop(handle)
+
+    def _get_callbacks_for_event_type(
+        self, event_type: EventType
+    ) -> Iterator[Callable[[Event], None]]:
+        """Returns an iterator to all callbacks associated with th eenevent type
```
Lietze: Typo
```suggestion
        """Returns an iterator to all callbacks associated with the relevant type
```

src/logger/logger.py
```diff
@@ -21,470 +24,521 @@
...
-        :param tick: The current simulation tick
-        :param train_speed_fault_configuration: The configuration of the fault
-        :param affected_element: The affected element
-        :param value_before: The value before the fault
-        :param value_after: The value after the fault
-        :rtype: None
+        :param event: the event containing all relevant info
         """
+        if "train_speed_fault_configuration" not in event.arguments:
+            return
```
Lietze, line 412: Why did you add this? I think that should not ever happen right?

src/logger/logger.py
```diff
@@ -21,470 +24,521 @@
...
         """
         This function should be called when removing a platform blocked fault from the simulation.
         This should reference the fault configuration of the fault.
-        :param tick: The current simulation tick
-        :param platform_blocked_fault_configuration: The configuration of the fault
-        :rtype: None
+        :param event: the event containing all relevant info
         """
+        if "platform_blocked_fault_configuration" not in event.arguments:
+            return
```
Lietze, line 436: Same here. Does this ever happen?

Lietze: Also everywhere else, I'm not writing the same comment 20 times ;)

src/event_bus/event_bus.py
```diff
@@ -0,0 +1,202 @@
...
+        defined in EVENT_METHODS. It returns a callable that creates
+        an event and calls all callbacks.
+
+        :param name: the method name
+        :raises AttributeError: when unknown method name is passed
+        :return: a callable
+        """
+        if name not in self.EVENT_METHODS:
+            raise AttributeError(
+                f"'{self.__class__.__name__}' object has no attribute '{name}'"
```
Lietze, line 197: Why does self. __class__. __name__ not always refer to EventBus? If it does, I think you should make it explicit.

src/event_bus/event_bus.py
```diff
@@ -0,0 +1,202 @@
...
+            )
+        arg_keys, event_type = self.EVENT_METHODS[name]
+
+        def emit_event(*args: list, **kwargs: dict):
+            arguments = self._build_arguments(name, arg_keys, args, kwargs)
+            event = Event(event_type, arguments)
+            for callback in self._get_callbacks_for_event_type(event_type):
+                callback(event)
+
+        return emit_event
```
Lietze, line 207: How exacly does __getattr__ work and when is it called?
Is it called whenever any method or attribute of this class is called to find it and you are overwriting this?
And why does the return value get executed in the other components?

src/logger/logger.py
```diff
@@ -21,470 +24,521 @@
...
     run_id: UUID
+    callback_handles: list[UUID]
 
-    def __init__(self, run_id: UUID):
+    def __init__(self, run_id: UUID, event_bus: EventBus):
         """
         The constructor of the logger class
         """
+        super().__init__(event_bus, 5)
         self.run_id = run_id
```
Lietze: Does the Logger still need run_id for anything or could it be removed? I was wondering, because the EventBus also has run_id.

tests/fault_injector/conftest.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 from traci import vehicle
 
+from src.event_bus.event_bus import EventBus
 from src.implementor.models import Run, SimulationConfiguration, Token
 from src.interlocking_component.route_controller import IInterlockingDisruptor
 from src.logger.logger import Logger
```
Lietze, line 7: How come you still need the Logger?

tests/fault_injector/test_affected_element_does_not_exist.py
```diff
@@ -5,6 +5,7 @@
 import pytest
 
 from src.base_model import BaseModel
+from src.event_bus.event_bus import EventBus
 from src.fault_injector.fault_configurations.fault_configuration import (
     FaultConfiguration,
 )
```
Lietze, line 7: Further down there is still the Logger as an import. Do you still need that?

tests/fault_injector/test_platform_blocked_fault.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 
+from src.event_bus.event_bus import EventBus
 from src.fault_injector.fault_configurations.platform_blocked_fault_configuration import (
     PlatformBlockedFaultConfiguration,
 )
```
Lietze, line 6: Logger is still there as an import.

tests/fault_injector/test_schedule_blocked_fault.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 
+from src.event_bus.event_bus import EventBus
 from src.fault_injector.fault_configurations.schedule_blocked_fault_configuration import (
     ScheduleBlockedFaultConfiguration,
 )
```
Lietze, line 6: Logger as Import

tests/fixtures/fixtures_spawner.py
```diff
@@ -757,15 +759,22 @@ def let_next_spawn_fail(self):
     return MockTrainSpawner()
 
 
+@pytest.fixture
+def event_bus(run) -> EventBus:
+    bus = EventBus(run_id=run.id)
+    Logger(run.id, bus)
+    return bus
```
Lietze: This is the same fixture as in fixtures/fixtures_logger.py. Why do you need it again?

tests/interlocking_component/conftest.py
```diff
@@ -137,10 +137,10 @@ def get_gg_count() -> int:
 
 @pytest.fixture
 def route_controller(
-    configured_souc: SimulationObjectUpdatingComponent, mock_logger: Logger
+    configured_souc: SimulationObjectUpdatingComponent, event_bus: EventBus
```
Lietze: Shouldn't this be different?
```suggestion
    configured_souc: SimulationObjectUpdatingComponent, mock_event_bus: EventBus
```

Lietze: Also apply any changes two lines down.

tests/logger/test_log_collector.py
```diff
@@ -224,342 +225,348 @@ def _enter_leave_block_section_all_df(self):
...
+        self.setup_departure_arrival_1(event_bus)
         assert_frame_equal(
-            log_collector.get_departures_arrivals_all_trains(logger.run_id),
+            log_collector.get_departures_arrivals_all_trains(event_bus.run_id),
             _departure_arrival_all_1_df,
         )
 
-    def test_departure_arrival_empty(self, logger: Logger, log_collector: LogCollector):
+    def test_departure_arrival_empty(
+        self, event_bus: Logger, log_collector: LogCollector
```
Lietze: Why is it `event_bus: Logger`? Does this make the typehints better or is there a real reason?

tests/logger/test_log_collector.py
```diff
@@ -581,5 +588,7 @@ def test_get_faults(
         faults_log_collector_df["fault_id"] = faults_log_collector_df[
             "fault_id"
         ].astype("string")
-        generated_faults_df = log_collector.get_faults(logger.run_id)
+        generated_faults_df = log_collector.get_faults(event_bus.run_id)
+        print(generated_faults_df)
+        print(faults_log_collector_df)
```
Lietze: Please remove these two lines.
```suggestion
```

src/logger/logger.py
```diff
@@ -21,470 +24,521 @@
...
+        :param event: the event containing all relevant info
         """
         TrainLeaveBlockSectionLogEntry.create(
             timestamp=datetime.now(),
-            tick=tick,
-            message=f"Train with ID {train_id} left block section with ID {block_section_id}",
+            tick=event.arguments["tick"],
+            message=f"Train with ID {event.arguments['train_id']} left block "
+            f"section with ID {event.arguments['block_section_id']} "
+            f"with length {event.arguments['block_section_length']}",
```
Lietze: Why did you add to the message?

tests/event_bus/test_event_bus.py
```diff
@@ -0,0 +1,103 @@
...
+    def test_event(
+        self, event_bus: EventBus, event_type: EventType, tick: int, train_id: str
+    ):
+        self.callback_called = False
+
+        def callback(event: Event):
+            self.callback_called = True
+            assert event.event_type == event_type
+            assert event.arguments["tick"] == tick
+            assert event.arguments["train_id"] == train_id
```
Lietze, line 65: Is there an way to make the asserts outside of the callback? I think that would be easier to understand.

Lietze (changes requested):
This looks great. I have a few questions and I fould a few things you may have overlooked.

src/logger/logger.py
instant-sky: This can be the case, because there are only the EventTypes `INJECT_FAULT` and `RESOLVE_FAULT`. This results in all of the specific faults have to check, if they are meant by the call or it addresses another fault type.

src/logger/logger.py
instant-sky: see above

src/event_bus/event_bus.py
CR1337: This version minimizes the amount of places where you would have to change code when you would rename the class. It therefore reduces redundancy. But if it really bugs you I can change it and rely on modern IDE magic for renaming :)

src/event_bus/event_bus.py
CR1337: `__getattr__` gets called when the caller tries to acces an attribute that doesn't exist. It then creates this attribute and returns it. Because the caller here tries to execute a method, `__getattr__` has to return a callable that then gets executed by the caller.

tests/logger/test_log_collector.py
CR1337: We just overlooked this. It will be changed :)

tests/event_bus/test_event_bus.py
CR1337: I don't know how. I want to test if the argument passed to the callback contains the correct data. 

Do you have an idea how to test it otside the callback?

src/event_bus/event_bus.py
Lietze: No this makes sense to me. God job

src/event_bus/event_bus.py
Lietze: All right. This feels very efficient but also very complicated at the same time. I think it is a good way of doing it, but maybe you could add some comments to prevent some headaches.

tests/event_bus/test_event_bus.py
Lietze: No. This is also fine.

src/event_bus/event_bus.py
CR1337: I didn't want to create a method for every possible event which would effectively all do the same.

But you are right. I will add some comments to clarify the functionality.

src/logger/logger.py
CR1337: You are right. I removed it. Thanks :)

tests/fault_injector/conftest.py
CR1337: Prior to this refactoring tests using fixtures from this file used the `Logger` and expected `LogEntry`s to be created. Now `Components` have the `EventBus` instead of the `Logger`. Nonetheless `LogEntry`s still need to be created. Therefore the Logger still needs to be instantiated here.

tests/fault_injector/test_affected_element_does_not_exist.py
CR1337: I will remove it.

tests/fault_injector/test_platform_blocked_fault.py
CR1337: Will be removed.

tests/fault_injector/test_schedule_blocked_fault.py
CR1337: Will be removed.

tests/fixtures/fixtures_spawner.py
CR1337: There was already some duplication prior to this refactoring. The duplicate fixtures partly rely on different other fixtures. If I simply remove the duplicate fixtures it breaks a number of tests. This should definitely be taken care of but I suggest doing it in a future issue due to the complexity this problem seems to have.

src/logger/logger.py
CR1337: I will change it back. The length is stored in the `LogEntry` anyway.

tests/data_science/test_data_science.py
```diff
@@ -136,70 +141,66 @@ def test_get_coal_demand_by_run_id(
...
-        data_science: DataScience,
-        window_size_time_by_config_id_df: pd.DataFrame,
-    ):
-        setup_logs_departure_arrival(logger)
-        setup_logs_departure_arrival_alt(logger2)
-        window_size_df = data_science.get_window_size_time_by_config_id(
-            simulation_configuration
-        )
-        assert_frame_equal(window_size_df, window_size_time_by_config_id_df)
-
```
Lietze, line 194: Why did you just remove this test?

tests/data_science/test_data_science.py
```diff
@@ -255,13 +256,13 @@ def test_get_coal_spawn_events_by_config_id(
     def test_get_window_by_config_id(
         self,
         simulation_configuration: SimulationConfiguration,
-        logger: Logger,
-        logger2: Logger,
+        event_bus: EventBus,
+        event_bus2: Logger,
```
Lietze: Why is this a Logger?

Lietze (commented):
I was wondering why you deleted the test from the data_science module? Also I think in a lot of tests there is still a Logger being used when it should be a EventBus. You should try to catch all of those.

tests/fixtures/fixtures_spawner.py
Lietze: That seems good.

tests/data_science/test_data_science.py
CR1337: The method that was tested was removed during development on this branch. After merging dev into here I had to delete this test manually because GitHub wasn't able to.

tests/data_science/test_data_science.py
CR1337: The `logger` here is passed to `setup_logs_departure_arrival` and `setup_logs_departure_arrival_alt`. Both prepare the `TestLogCollector` which directly interacts with the Logger. Therefore it needs a `Logger` to be passed as argument.

tests/data_science/test_data_science.py
Lietze: `get_window_size_time_by_config_id` does still exist in data_science.py in line 397. Or did you mean something else?

tests/data_science/test_data_science.py
Lietze: If `setup_logs_departure_arrival` and `setup_logs_departure_arrival_alt` both prepare the TestLogCollector, why does one get a Logger and the other a EventBus? If the fixture of the EventBus is passed, then it should just execute the callback to the logger anyway right?
What I'm not sure about is, if the data_science fixture would still have access to the created logs.

Lietze (approved):
Very nice, thank you!

tests/data_science/test_data_science.py
CR1337: I changed the type hints to `EventBus` now.

tests/data_science/test_data_science.py
CR1337: I don't know what went wrong here. I reenabled the test.

PR #523, opened 2023-05-30
Feature/518 split docker compose

### LINKED ISSUE #518: Split Docker Compose
## User stories
As a dev I want to split docker compose files in dev test pro in order to deploy the app and use env files better

## Acceptance Criteria
- [x] split docker compose in 3 files
- [x] split env files

LucasDerReisende:
Fixes #518 

Split docker compose files

## PR checklist


.env.dev
antonykamp, line 1: Please update our wiki 😇

.env.test
```diff
@@ -2,16 +2,18 @@
 DATABASE_NAME=postgres
 DATABASE_USER=postgres
 DATABASE_PASSWORD=root
+POSTGRES_PASSWORD=root
 DATABASE_HOST=localhost
-DATABASE_PORT=5430
+DATABASE_PORT=5432
```
antonykamp, line 7: I'm personally a fan of keeping the database port at 5430. Otherwise, you can't execute the tests while running the application. That's annoying sometimes 🙃 What do you think about it?

docker-compose.yaml
```diff
@@ -1,87 +1,35 @@
...
-    env_file:
-      - .env.shared
-    environment:
-      - GRAFANA_HOST=grafana-dev
-    depends_on:
-      - postgresql
-      - celery-worker
-      - grafana-dev
-
   postgresql:
```
antonykamp, line 4: That means we don't use the optimized version of the Postgres database in the tests anymore, is that correct? 😇

docker-compose.test.yaml
```diff
@@ -0,0 +1,15 @@
+version: "3.8"
+
+services:
+  postgresql:
+    command: "postgres -c 'shared_buffers=128MB' -c 'fsync=off' -c 'synchronous_commit=off' -c 'full_page_writes=off' -c 'max_connections=100' -c 'client_min_messages=warning'"
+    env_file:
+      - .env.test
+
+  celery-worker:
```
antonykamp, line 9: We won't use the celery-worker in the test cases because fixtures by celery will replace it. That means we'll never start the celery-worker container when we want to run tests. Do you know if we still need it here?

pyproject.toml
```diff
@@ -134,7 +134,7 @@ build-backend = "poetry.core.masonry.api"
 
     [tool.poe.tasks.test-docker-deps]
     help = "Initializes needed docker containers for tests"
-    cmd = "docker compose up -d grafana-test redis postgresql-test --wait"
+    cmd = "docker compose -f docker-compose.yaml -f docker-compose.test.yaml up -d grafana redis postgresql --wait"
```
antonykamp, line 137: Just out of curiosity: Do we have to specify `grafana redis postgres` anymore if we always want to start everything of the docker-compose file? I'm just interested 😇

antonykamp (commented):
Just having some questions to understand everything 😇

.env.test
LucasDerReisende: since it is now the same database, the ports will be overriden, so when running tests, the dev environment will stop and vice verca

docker-compose.yaml
LucasDerReisende: we still do, if you have a look at docker-compose.test.yaml all the parameters from postgres-test are overridden with this compose file

docker-compose.test.yaml
LucasDerReisende: can remove it

pyproject.toml
LucasDerReisende: yes, because, we are not using celery-worker, so it does not need to be started

.env.test
LucasDerReisende: It makes no difference switching the port, because the other instance would not be running anymore

.env.dev
LucasDerReisende: Which parts, I just added an entry to complete the description of env Variables

antonykamp: nvm thanks 😇

docker-compose.yaml
antonykamp: Sorry and thanks 😇 It's already late :P

.env.test
antonykamp: I don't completely understand 🤔 Can we speak tomorrow please 😇

LucasDerReisende: Yes sure :)

LucasDerReisende: Before we had two different postgresql containers (dev and test), now it is one for both
But the environment variables and start command will be changed if we start it in a different environment

antonykamp: Ah thanks :D Now I got it.

antonykamp (approved):
Go for it 🥳

PR #532, opened 2023-06-01
Bug/531 skipping trains

### LINKED ISSUE #531: When having a short track segment and a fast train, the train skipps emitting the drove onto/off track event
## Bug: train skips short track segments

The train shouldn't skip the segment

Potential fix:
Increment the ticks per second

```
We want to drive 100 km/h
We want to trigger a potential update roughly every 5-10m
100km/h = 27.7m/s
27.7m/s / 5 = 5.5
```
We need roughly 6 ticks every second to capture a data-point every 5m. We should probably just capture every 100ms / 200ms for round numbers

SaturnHafen:
Fixes #531

## PR checklist


SaturnHafen:
@LucasDerReisende @CR1337 @Lietze you all said that your code depends on 1 tick = 1 second. Do you have any pointers where I should start fixing? (Or will you fix these issues for yourself?)

I added the `TICK_LENGTH` env var to declare the length of a tick. To get the ticks per second, you can calculate `1 / TICK_LENGTH`.

CR1337:
> @LucasDerReisende @CR1337 @Lietze you all said that your code depends on 1 tick = 1 second. Do you have any pointers where I should start fixing? (Or will you fix these issues for yourself?)
> 
> I added the `TICK_LENGTH` env var to declare the length of a tick. To get the ticks per second, you can calculate `1 / TICK_LENGTH`.

Before changing any code I need to know whether the schedules should use ticks or seconds. In other words: Should the user tell the system that a schedule should start at `x seconds` or at `tick y`?

LucasDerReisende:
> @LucasDerReisende @CR1337 @Lietze you all said that your code depends on 1 tick = 1 second. Do you have any pointers where I should start fixing? (Or will you fix these issues for yourself?)
> 
> I added the `TICK_LENGTH` env var to declare the length of a tick. To get the ticks per second, you can calculate `1 / TICK_LENGTH`.

I will make some changes in the tests and evaluate what needs to be changed

SaturnHafen:
> Before changing any code I need to know whether the schedules should use ticks or seconds. In other words: Should the user tell the system that a schedule should start at `x seconds` or at `tick y`?

My preference would be that the schedule considers seconds. I don't think we need sub-second precision. But we should discuss this in the weekly.

LucasDerReisende:
I did some changes to the data science and logger code. Some tests will still fail, because before I I can fix those, the Schedule Code needs to be updated. I would be grateful if you could ping me, when this has been done :)

CR1337:
I currently change the `tick`-based inner workings of the `Schedule` to `second`-based. Should the train_id contain the `tick` when the train should be spawned or the amount of elapsed `seconds` when the train should be spawned?
@SaturnHafen @LucasDerReisende

CR1337:
Except for the thing with the train id I am done with my part now.

LucasDerReisende:
> I currently change the `tick`-based inner workings of the `Schedule` to `second`-based. Should the train_id contain the `tick` when the train should be spawned or the amount of elapsed `seconds` when the train should be spawned? @SaturnHafen @LucasDerReisende

I do not have a strong opinion on this, but would opt for using ticks, since the train_id is part of the inner workings of the simulation (thus should use ticks) and it would also not interfere, when somebody is using the train_id for gaining the start_tick. I am not doing this, but I need the train_id to get the type of train from the logs.
On the other hand you could argue, that train names, when displayed in the dashboard are somewhat longer,

CR1337:
> > I currently change the `tick`-based inner workings of the `Schedule` to `second`-based. Should the train_id contain the `tick` when the train should be spawned or the amount of elapsed `seconds` when the train should be spawned? @SaturnHafen @LucasDerReisende
> 
> I do not have a strong opinion on this, but would opt for using ticks, since the train_id is part of the inner workings of the simulation (thus should use ticks) and it would also not interfere, when somebody is using the train_id for gaining the start_tick. I am not doing this, but I need the train_id to get the type of train from the logs. On the other hand you could argue, that train names, when displayed in the dashboard are somewhat longer,

Ok, the train ids now contain the ticks.

LucasDerReisende:
From my side everything should also be updated

LucasDerReisende:
All Data Science functions should also be updated to work with an adaptive tick length

antonykamp (approved):
Sorry for the late review 😇

PR #543, opened 2023-06-09
Give components correct priorities

### LINKED ISSUE #541: Give components correct priorities
## User stories
As a user I the components to be called in the correct order in order to make sure that when calling a component nothing from other components is missing.

currently planned order:
Updater -> FualtInjector -> RouteController -> Rest

## Acceptance Criteria
- [ ] give every component the correct priority

instant-sky:
Fixes #541 

## PR checklist

### DIFF
--- a/src/component.py
+++ b/src/component.py
@@ -16,7 +16,7 @@ class Component(ABC):
         "LOW": 7,
     }
 
-    def __init__(self, event_bus: EventBus, priority: str):
+    def __init__(self, event_bus: EventBus, priority: str = "LOW"):
         """Initializes the component.
 
         :param event_bus: reference to the global event_bus
@@ -39,7 +39,7 @@ class MockComponent(Component):
     """Mock for a simple component to check if next tick is called"""
 
     def __init__(self):
-        Component.__init__(self, None, 1)
+        Component.__init__(self, None)
 
     def next_tick(self, tick: int):
         pass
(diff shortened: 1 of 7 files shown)


antonykamp (commented):
Thanks for the change 👍 
I have one question and one idea 😇 What priority does the SimulationObjectUpdater have, and what do you think about an Enum for priorities instead of magic numbers 🙈

instant-sky:
The `SimulationObjectUpdatingComponent` has priority 10. It was already like this so I did not have to add it by myself. 

An enum could be an option. Would you name the priorities something like `VERY_HIGH`, `HIGH`, `MEDIUM`, `LOW` or maybe something like `FIRST`, `SECOND`, `THIRD`, ...? 🤔

antonykamp:
> An enum could be an option. Would you name the priorities something like VERY_HIGH, HIGH, MEDIUM, LOW or maybe something like FIRST, SECOND, THIRD, ...? 🤔

I like your first proposal 😇

src/component.py
```diff
@@ -9,14 +9,21 @@ class Component(ABC):
     event_bus: EventBus
     priority: int
 
-    def __init__(self, event_bus: EventBus, priority: int):
+    PRIORITY_LEVELS: dict[str, int] = {
```
antonykamp, line 12: I thought about real enums, actually 😇 We could compare them and use them as types. But if you are happy with this solution, then I am too.

antonykamp (approved):

src/component.py
instant-sky: I think it is sufficient for now. But we can change that in the future if it is necessary? 🤔

PR #553, opened 2023-06-23
moved and rename enter and leave block_section

### LINKED ISSUE #552: Fix Train Enter Block Section and Train Leave Block Section
## User stories
As a developer I want to have the correct train enter/leave log entries in order to get accurate results in the dashboard

## Acceptance Criteria
- [ ] remove train enter/leave block section calls from route controller
- [ ] add correct train enter/leave block section calls to infrastructure provide
- [ ] rename method to enter_edge and leave_edge instead of block_section

Lietze:
Fixes #552 

## PR checklist


src/interlocking_component/route_controller.py
LucasDerReisende, line 1: Line 354/345: station_index attribute does not exist on uninitialized train

LucasDerReisende, line 1: Line 296/287: type mismatch: expect train, got uninitialized train

src/interlocking_component/route_controller.py
```diff
@@ -243,20 +243,12 @@ def __init__(
 
     def next_tick(self, tick: int):
         self.tick = tick
-        for (
-            interlocking_route,
-            train,
-            route_length,
-        ) in self.route_queues.routes_to_be_set:
+        for (interlocking_route,) in self.route_queues.routes_to_be_set:
```
LucasDerReisende: remove tuple unwrapping

src/interlocking_component/route_controller.py
```diff
@@ -243,20 +243,12 @@ def __init__(
...
             # The Sumo route was already set and the route was reserved.
-            was_set = self.set_interlocking_route(
-                interlocking_route, train, route_length
-            )
+            was_set = self.set_interlocking_route(interlocking_route)
             if was_set:
-                self.route_queues.routes_to_be_set.remove(
-                    (interlocking_route, train, route_length)
-                )
+                self.route_queues.routes_to_be_set.remove((interlocking_route))
```
LucasDerReisende: braces can be removed

src/interlocking_component/route_controller.py
```diff
@@ -417,19 +405,13 @@ def set_fahrstrasse_if_reservations_work(
...
         if self.check_if_route_is_reserved_as_first(route, train, entire_route):
-            was_set = self.set_interlocking_route(
-                interlocking_route, train, route_length
-            )
+            was_set = self.set_interlocking_route(interlocking_route)
             if not was_set:
-                self.route_queues.routes_to_be_set.append(
-                    (interlocking_route, train, route_length)
-                )
+                self.route_queues.routes_to_be_set.append((interlocking_route))
```
LucasDerReisende: braces can be removed

src/interlocking_component/route_controller.py
```diff
@@ -243,20 +243,12 @@ def __init__(
...
-            was_set = self.set_interlocking_route(
-                interlocking_route, train, route_length
-            )
+            was_set = self.set_interlocking_route(interlocking_route)
             if was_set:
-                self.route_queues.routes_to_be_set.remove(
-                    (interlocking_route, train, route_length)
-                )
+                self.route_queues.routes_to_be_set.remove((interlocking_route))
         for route, train in self.route_queues.routes_to_be_reserved:
```
LucasDerReisende: does this need to be route, train or just route?

LucasDerReisende (changes requested):
annotated some lines, that caused errors with me, maybe check and run a simulation for yourself to see, if there are any others

src/interlocking_component/route_controller.py
Lietze: It should be `route, train`

src/interlocking_component/route_controller.py
Lietze: Does this happen in a test? From where does the call come, when the type mismatch happen?

Lietze: This should be fixed in the next commit

src/interlocking_component/route_controller.py
Lietze: This should be fixed in the next commit

src/interlocking_component/route_controller.py
Lietze: This should be fixed in the next commit

src/interlocking_component/route_controller.py
Lietze: This should be fixed in the next commit

src/interlocking_component/route_controller.py
LucasDerReisende: When calling `self.set_fahrstrasse(train_to_be_initialized, starting_edge)` in `set_spawn_fahrstrasse`, the function `set_fahrstrasse` expects a train as type Train. But the object `train_to_be_initialized` is of Type `UninitializedTrain`.

Lietze: Ok, that is actually the expected behavior. I implemented `UninitializedTrain` as a Mock Object for the normal `Train` object, because it is not possible to use the train before it is initialized, but it needs the first route before that. I thought that python does not care about what Type an object is as long, as it implements the methods that are being called.
Do you know how to prevent that error from being thrown without changing the functionality?

Lietze: Or is this a linting issue?

LucasDerReisende (approved):
looks good

PR #567, opened 2023-07-12
redundant queues

### LINKED ISSUE #563: redundant queues for routes to be reserved
## User stories
As a developer I want information to not be redundant in order to save time and memory

## Acceptance Criteria
- [ ] routes_to_be_reserved or routes_waiting_for_reservations is removed
- [ ] the timing of reserving is changed
    - [ ] after the SUMO-Route is set
    - [ ] before interlocking route is set

Lietze:
Fixes #563

## PR checklist

### DIFF
--- a/src/interlocking_component/route_controller.py
+++ b/src/interlocking_component/route_controller.py
@@ -196,14 +196,12 @@ class RouteController(Component):
         """This class capsules all routes, that need to be considered every tick."""
 
         routes_to_be_set: List[Tuple[Route, Train, int]]
-        routes_to_be_reserved: List[Tuple[Route, Train]]
         routes_waiting_for_reservations: List[
             Tuple[List[Node], Train, Route, float, List[Node]]
         ]
 
         def __init__(self) -> None:
             self.routes_to_be_set = []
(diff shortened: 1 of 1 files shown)


antonykamp (commented):
Why don't you need this queue anymore? 😇 Does the logic never apply, or does another list the same job?

Lietze:
> Why don't you need this queue anymore? 😇 Does the logic never apply, or does another list the same job?

There are two reasons a queue like that could be needed for reservations. The first is whether the reservation itself is possible. If not, next tick that needs to be tried again. The other is if the SUMO-Route, the train will drive on is reserved in first position for the train. If that is not the case yet, it needs to wait. By moving the reserving of the route right in front of the check if it is reserved as first, both of those cases can now be handled by the same queue.
This has the downside, that if a train is waiting for it's reservation to be first on the route, each tick it will be checked, if the route is reserved at all. Two seperate queues prevented that. however I don't think the impact on performance is noticable and I want the code to look better.

antonykamp (approved):
Thanks for the detailed explanation and your time! It helped allot and I like you changes! 💚

PR #568, opened 2023-07-12
reservations swapping eventhough the train is already driving towards it

### LINKED ISSUE #558: Reservation swaping in used the block section
## Problem Description

Reservations can be swaped even though a train is already in the block section where the reservations may be swapped

1. Reservations are swapped although other train is on that blocksection
2. This should not be passile. It should only be able to swap if the other train is not in that blocksection yet
3. It messes with the assumtion, that a train always drives over a track, that is reserved for it

Problematic code:
```
if edge.track.reservations[0][0].edge == edge:
    # The train this is reserved for is already on the swgment, that may be swaped.
    return
```

## Steps For Reproduction

1. One train drives into block section
2. Another trin wants to cross through that block section, when first train is not on the reservation track
3. Other trains reservations are put as first

Lietze:
Fixes #558 

## PR checklist

### DIFF
--- a/src/interlocking_component/route_controller.py
+++ b/src/interlocking_component/route_controller.py
@@ -609,8 +609,9 @@ class RouteController(Component):
         for i, edge in enumerate(edge_route):
             if not edge.track.is_reservation_track:
                 continue
-            if edge.track.reservations[0][0].edge == edge:
-                # The train this is reserved for is already on the swgment, that may be swaped.
+            first_reserving_train = edge.track.reservations[0][0]
+            if edge in first_reserving_train.reserved_tracks[0].edges:
+                # The train this is reserved for is already driving towards the segment, that may be swaped.
                 return
             if edge.track.reservations[0][0] == train:
                 self.put_reservations_as_first(train, edge_route[:i])
--- a/src/interlocking_component/route_controller.py
+++ b/src/interlocking_component/route_controller.py
@@ -611,7 +611,8 @@ class RouteController(Component):
                 continue
             first_reserving_train = edge.track.reservations[0][0]
             if edge in first_reserving_train.reserved_tracks[0].edges:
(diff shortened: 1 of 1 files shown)


src/interlocking_component/route_controller.py
```diff
@@ -609,8 +609,10 @@ def maybe_put_reservations_as_first(self, train: Train, route: List[Node]):
         for i, edge in enumerate(edge_route):
             if not edge.track.is_reservation_track:
                 continue
-            if edge.track.reservations[0][0].edge == edge:
-                # The train this is reserved for is already on the swgment, that may be swaped.
+            first_reserving_train = edge.track.reservations[0][0]
```
antonykamp, line 612: Does this part check for other trains in the block section too?

Lietze: Not exactly. It checks if the train is on the `ReservationTrack` or close to it. Close in this context means, that this `ReservationTrack` is the next `ReservationTrack`, the train has an reservation for. Each block section apart from those crossing switches has exactly one `ReservationTrack`. If the train is not on it's previous `ReservationTrack`, the check for this train, if it has the neccassary reservations, has checked the `ReservationTrack` that may be swaped and because of that has allowed the train to continue driving. If the reservation is then swaped away from first position, the train may drive toward a `ReservationTrack`, that is not reserved for it in first position. This will lead to a situation, where the first reserving train is stuck behind the second reserving train.
To prevent that we need to make sure, that prior to swaping reservations the first reserving train is still on or driving towards another `ReservationTrack`. If the reservations get swapped then, this train will not be allowed to drive towards the `ReservationTrack` where reservations were swapped, because it does not have the reservation in first position there.
It is not importent, if a train is on the same blocksection after the `ReservationTrack`, because it can always just leave and crashes are prevented by the use of the interlocking.

antonykamp (approved):

